PR #6, opened 2018-06-25
issue 5: Introduced transactions per commit.

### LINKED ISSUE #5: Transactions not committing with the tool ..
Transactions not committing with the tool .. I don't see any activity on the segment heat map or throughput  during the transactions test with 30 producers and 10k events per sec ..

arvindkandhare:
Signed-off-by: arvindkandhare <arvind.kandhare@emc.com>

**Change log description**
Add parameter which decides number of events per commit of a transaction.

**Purpose of the change**
Fixes #5.

**What the code does**
Adds a new parameter `transactionspercommit`. A transaction will be committed once `transactionpercommit` number of events are sent on it and a new transaction is created.

**How to verify it**
Verified against `standalone`.


ack72tdp:
Tried this code .. and it seem to work fine .. I can see commits happening in the dashboard ...

PR #10, opened 2018-09-19
Manual scaling, runtime, latency and segments improvements

kmgowda:
1.	Runtime improvement
•	The time is stamped between the starting and ending of the number of events per sec. if it has taken more time then reduce the iterations to complete the bench-marking with in the time
•	In case if the number of events completes with in second, then thread should wait till remaining time to cross 1 second.

2.	Latency accuracy
•	The time is stamped before and after the write event.
•	In case of blocking writes, the time is stamped before and after the write event and ack

3.	Segments options
•	If the existing segment size is not matching with number of producers then its updated
•	The new option “-segments ”, creates the number of segments which can less than or more than number of producers.
•	The option “-segments ” not supplied then number of segments is set to number of producers

4.	Fork/join option
•	The fork executor service is used to improve the number of processors usage. If the latency is low and higher throughput condition, the execution of the parallel thread improve in chronograph. This condition tries to write more data in parallel.

5.	Manual Scaling
•	If the stream is already existing then stream is scaling input number of segments; by default, the number of producers are set as number of segments
•	If the “-recreate true” is provided, then instead of manual scaling, the existing stream will be sealed, deleted and recreated.

### DIFF
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -21,6 +21,7 @@ package com.emc.pravega.perf;
 
 
 import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.function.Supplier;
@@ -31,9 +32,8 @@ class PerfStats {
     private long windowStartTime;
     private long start;
     private long windowStart;
-    private double[] latencies;
+    private ArrayList<Double> latencies;
     private int iteration;
-    private int index;
     private long count;
     private long bytes;
     private double maxLatency;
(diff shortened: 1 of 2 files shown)


arvindkandhare:
> 2\. The time is stamped before and after the write event.

I do not think this is a correct strategy. Latency is the time between event and ack. Not the time taken by the write call to return.

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -53,59 +50,48 @@ public PerfStats(String action, long numRecords, int reportingInterval, int mess
             this.windowStart = 0;
             this.index = 0;
             this.iteration = 0;
-            this.sampling = (int) (numRecords / Math.min(numRecords, 500000));
-            this.latencies = new long[(int) (numRecords / this.sampling)];
-            this.index = 0;
+            this.latencies = new double[(int) (numRecords / reportingInterval)];
```
arvindkandhare: Can you explain the change? IMO we will lose track of the earlier latencies and use only final reporting window to report overall latency. This may not show a true picture.

kmgowda: hi Arvind
   Thanks for your review comments..
This is written with intention that my sampling time is same as reporting time... 
for example, if i have set the runtime as 1 minute and reporting interval is 1 sec, for every 1 sec i will collect the latency sample..   so , in 1 minute , I will collect the 60 samples if the reporting interval is set to 1 sec...

kmgowda:
> > 2. The time is stamped before and after the write event.
> 
> I do not think this is a correct strategy. Latency is the time between event and ack. Not the time taken by the write call to return.

Hi Arvind
     
1. The time is stamped before and after the "write event"  -- for non blocking writes
2. the time is stamped before and after the "write event and ack"  --- for blocking writes
           This change is already there in the PR.



Here is code snippet to record the time and check the blocking or non blocking

   public CompletableFuture writeAndRecordTime(Supplier<CompletableFuture> fn, int length, boolean blocking ) {
        CompletableFuture  retVal=null;
        long startTime, endTime;
        startTime = System.currentTimeMillis();
        try {
             retVal = fn.get();
             if (blocking) {
                  retVal.get();
             }
        } catch (Exception e) {
             e.printStackTrace();
        }
        endTime = System.currentTimeMillis();
        
        if(retVal == null) {
            record(length, startTime,  endTime);
        } else {
            retVal = retVal.thenAccept((d) -> {
                record(length, startTime, endTime);
            });
        }
        return retVal;
    }

kmgowda:
hi Arvind
   This PR was raised on the master of my fork, I could not able to send the update on this master.
so, I have created another PR : https://github.com/pravega/pravega-benchmark/pull/11
I have covered your review comments about the latecy measurement. please let me know if you any further comments

PR #11, opened 2018-09-27
latency, manual scaling, runtime and  segments improvements

kmgowda:
This PR is updated with following changes

1. boolean returns are replaced with Exception
2. code cleanup as per Tom's review comments

This PR  Covered Tom's review comments with the following changes
1. The Runnable classes are changed to callable to return the exceptions
       to caller or parent thread.
2. the boolean values and other user inputs are handled at main
3. in case of exception, its stacked back to main and exit; so that its final and single of exit for the benchmarking
4. the "blocking" option is removed as it not required; 
5. the fork join is made the default and thread executor service is removed to improve the cpu usage 
6. the code is refactor to use OO concepts and streams, callables , fork join parallelism effectively

Covered Arvind's review comments on https://github.com/pravega/pravega-benchmark/pull/10

Runtime improvement
•	The time is stamped between the starting and ending of the number of events per sec. if it has taken more time then reduce the iterations to complete the bench-marking with in the time
•	In case if the number of events completes with in second, then thread should wait till remaining time to cross 1 second.

Latency accuracy
•	The start time is stamped before the write event and end time is stamped in the callback thread of the ack of the completable future of write.
•	In case of blocking writes, the time is stamped before and after the write event and ack

Segments options
•	If the existing segment size is not matching with number of producers then its updated
•	The new option “-segments ”, creates the number of segments which can less than or more than number of producers.
•	The option “-segments ” not supplied then number of segments is set to number of producers

Fork/join option
•	The fork executor service is used to improve the number of processors usage. If the latency is low and higher throughput condition, the execution of the parallel thread improve in chronograph. This condition tries to write more data in parallel.

Manual Scaling
•	If the stream is already existing then stream is scaling input number of segments; by default, the number of producers are set as number of segments
•	If the “-recreate true” is provided, then instead of manual scaling, the existing stream will be sealed, deleted and recreated.

### DIFF
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- * <p>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -140,18 +140,11 @@ class PerfStats {
 
     }
 
-    public CompletableFuture writeAndRecordTime(Supplier<CompletableFuture> fn, int length, boolean blocking ) {
+    public CompletableFuture writeAndRecordTime(Supplier<CompletableFuture> fn, int length) {
         CompletableFuture  retVal=null;
         final long startTime = System.currentTimeMillis();
  
-        try {
-             retVal = fn.get();
-             if (blocking) {
-                  retVal.get();
-             }
-        } catch (Exception e) {
-             e.printStackTrace();
-        }
+        retVal = fn.get();
         
         if(retVal == null) {
             final long endTime = System.currentTimeMillis(); 
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -24,11 +24,6 @@ import java.util.concurrent.Executor;
 import io.pravega.client.stream.TxnFailedException;
 import io.pravega.client.stream.ReinitializationRequiredException;
 
-@FunctionalInterface
-interface SupplierWithCE<T, X extends Exception> {
-    T get() throws X;
-}
-
 class PerfStats {
     final private int messageSize;
     final private String action;
@@ -84,7 +79,7 @@ class PerfStats {
         this.count++;
 
         System.out.printf("%8d records %s, %9.1f records/sec, %9.3f MB/sec, %7.4f ms avg latency.\n",
-                windowCount, action, recsPerSec, mbPerSec, latency);
+            windowCount, action, recsPerSec, mbPerSec, latency);
     }
 
     private void newWindow(long currentNumber) {
@@ -109,8 +104,8 @@ class PerfStats {
         double mbPerSec = 1000.0 * this.bytes / (double) elapsed / (1024.0 * 1024.0);
         //double[] percs = percentiles(this.latencies, 0.5, 0.95, 0.99, 0.999);
         System.out.printf(
-                "%d records %s, %.3f records/sec, %d bytes record size, %.3f MB/sec, %.4f ms avg latency, %.4f ms max latency\n",
-                iteration, action, recsPerSec, messageSize, mbPerSec, totalLatency / ((double) count), (double) maxLatency);
+            "%d records %s, %.3f records/sec, %d bytes record size, %.3f MB/sec, %.4f ms avg latency, %.4f ms max latency\n",
+            iteration, action, recsPerSec, messageSize, mbPerSec, totalLatency / ((double) count), (double) maxLatency);
         /*  
         System.out.printf("latencies percentiles:  %.4f ms 50th, %.4f ms 95th, %.4f ms 99th, %.4f ms 99.9th.\n",
                            percs[0], percs[1], percs[2], percs[3]);
@@ -128,26 +123,7 @@ class PerfStats {
         return values;
     }
 
-    public CompletableFuture runAndRecordTime(SupplierWithCE<CompletableFuture, ReinitializationRequiredException> fn, long startTime, int length) throws ReinitializationRequiredException {
-        CompletableFuture retVal = fn.get();
-        if (retVal == null) {
-            final long endTime = System.currentTimeMillis();
-            record(length, startTime, endTime);
-        } else {
-            retVal = retVal.thenAccept((d) -> {
-                final long endTime = System.currentTimeMillis();
-                record(length, startTime, endTime);
-            });
-        }
-        return retVal;
-    }
-
-    public CompletableFuture writeAndRecordTime(SupplierWithCE<CompletableFuture, TxnFailedException> fn, int length) throws TxnFailedException {
-        CompletableFuture retVal = null;
-        final long startTime = System.currentTimeMillis();
-
-        retVal = fn.get();
-
+    public CompletableFuture recordTime(CompletableFuture retVal, long startTime, int length) {
         if (retVal == null) {
             final long endTime = System.currentTimeMillis();
             record(length, startTime, endTime);
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -125,7 +125,7 @@ class PerfStats {
         return values;
     }
 
-    public CompletableFuture runAndRecordTime(Supplier<CompletableFuture> fn, long startTime, int length) {
+    public CompletableFuture runAndRecordTime(Supplier<CompletableFuture> fn, long startTime, int length) throws Exception {
         CompletableFuture  retVal = fn.get();
         if(retVal == null) {
             final long endTime = System.currentTimeMillis();
@@ -140,7 +140,7 @@ class PerfStats {
 
     }
 
-    public CompletableFuture writeAndRecordTime(Supplier<CompletableFuture> fn, int length) {
+    public CompletableFuture writeAndRecordTime(Supplier<CompletableFuture> fn, int length) throws Exception {
         CompletableFuture  retVal=null;
         final long startTime = System.currentTimeMillis();
  
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- * 
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -46,26 +46,26 @@ class PerfStats {
     final private long reportingInterval;
 
     public PerfStats(String action, int reportingInterval, int messageSize) {
-           this.action = action;
-           this.start = System.currentTimeMillis();
-           this.windowStartTime = System.currentTimeMillis();
-           this.windowStart = 0;
-           this.iteration = 0;
-           this.latencies = new  ArrayList<Double>();
-           this.maxLatency = 0;
-           this.totalLatency = 0;
-           this.windowCount = 0;
-           this.windowBytes = 0;
-           this.reportingInterval = reportingInterval;
-           this.messageSize = messageSize;
+        this.action = action;
+        this.start = System.currentTimeMillis();
+        this.windowStartTime = System.currentTimeMillis();
+        this.windowStart = 0;
+        this.iteration = 0;
+        this.latencies = new ArrayList<Double>();
+        this.maxLatency = 0;
+        this.totalLatency = 0;
+        this.windowCount = 0;
+        this.windowBytes = 0;
+        this.reportingInterval = reportingInterval;
+        this.messageSize = messageSize;
     }
 
     public synchronized void record(int bytes, long startTime, long endTime) {
         this.iteration++;
         this.windowBytes += bytes;
-        this.windowCount++;        
+        this.windowCount++;
         /* did we arrived at reporting time */
-        if ((endTime - windowStartTime)  >= reportingInterval) {
+        if ((endTime - windowStartTime) >= reportingInterval) {
             printWindow(endTime);
             newWindow(count);
         }
@@ -73,18 +73,18 @@ class PerfStats {
 
     private void printWindow(long endTime) {
         long elapsed = endTime - windowStartTime;
-        double latency=(double)(elapsed / (double) windowCount);
+        double latency = (double) (elapsed / (double) windowCount);
         double recsPerSec = 1000.0 * windowCount / (double) elapsed;
         double mbPerSec = 1000.0 * this.windowBytes / (double) elapsed / (1024.0 * 1024.0);
 
         this.bytes += this.windowBytes;
-        this.totalLatency += latency; 
-        this.maxLatency = Math.max(this.maxLatency, latency); 
+        this.totalLatency += latency;
+        this.maxLatency = Math.max(this.maxLatency, latency);
         this.latencies.add(latency);
         this.count++;
-        
+
         System.out.printf("%8d records %s, %9.1f records/sec, %9.3f MB/sec, %7.4f ms avg latency.\n",
-                           windowCount, action, recsPerSec, mbPerSec, latency);  
+                windowCount, action, recsPerSec, mbPerSec, latency);
     }
 
     private void newWindow(long currentNumber) {
@@ -109,17 +109,17 @@ class PerfStats {
         double mbPerSec = 1000.0 * this.bytes / (double) elapsed / (1024.0 * 1024.0);
         //double[] percs = percentiles(this.latencies, 0.5, 0.95, 0.99, 0.999);
         System.out.printf(
-                "%d records %s, %.3f records/sec, %d bytes record size, %.3f MB/sec, %.4f ms avg latency, %.4f ms max latency\n", 
-                iteration, action, recsPerSec,  messageSize, mbPerSec, totalLatency / ((double) count), (double) maxLatency);
+                "%d records %s, %.3f records/sec, %d bytes record size, %.3f MB/sec, %.4f ms avg latency, %.4f ms max latency\n",
+                iteration, action, recsPerSec, messageSize, mbPerSec, totalLatency / ((double) count), (double) maxLatency);
         /*  
         System.out.printf("latencies percentiles:  %.4f ms 50th, %.4f ms 95th, %.4f ms 99th, %.4f ms 99.9th.\n",
                            percs[0], percs[1], percs[2], percs[3]);
         */
-        
+
     }
 
     private double[] percentiles(double[] latencies, double... percentiles) {
-        Arrays.sort(latencies, 0, (int)count);
+        Arrays.sort(latencies, 0, (int) count);
         double[] values = new double[percentiles.length];
         for (int i = 0; i < percentiles.length; i++) {
             int index = (int) (percentiles[i] * count);
@@ -129,36 +129,34 @@ class PerfStats {
     }
 
     public CompletableFuture runAndRecordTime(SupplierWithCE<CompletableFuture, ReinitializationRequiredException> fn, long startTime, int length) throws ReinitializationRequiredException {
-        CompletableFuture  retVal = fn.get();
-        if(retVal == null) {
+        CompletableFuture retVal = fn.get();
+        if (retVal == null) {
             final long endTime = System.currentTimeMillis();
-            record(length,startTime,  endTime);
+            record(length, startTime, endTime);
         } else {
             retVal = retVal.thenAccept((d) -> {
                 final long endTime = System.currentTimeMillis();
-                record(length,startTime, endTime);
-            });
-        }
-        return retVal;
-
-    }
-
-    public CompletableFuture writeAndRecordTime(SupplierWithCE<CompletableFuture, TxnFailedException> fn, int length) throws TxnFailedException {
-        CompletableFuture  retVal=null;
-        final long startTime = System.currentTimeMillis();
-
-        retVal = fn.get();
-
-        if(retVal == null) {
-            final long endTime = System.currentTimeMillis(); 
-            record(length, startTime,  endTime);
-        } else {
-            retVal = retVal.thenAccept((d) -> {
-                final long endTime = System.currentTimeMillis(); 
                 record(length, startTime, endTime);
             });
         }
         return retVal;
     }
 
+    public CompletableFuture writeAndRecordTime(SupplierWithCE<CompletableFuture, TxnFailedException> fn, int length) throws TxnFailedException {
+        CompletableFuture retVal = null;
+        final long startTime = System.currentTimeMillis();
+
+        retVal = fn.get();
+
+        if (retVal == null) {
+            final long endTime = System.currentTimeMillis();
+            record(length, startTime, endTime);
+        } else {
+            retVal = retVal.thenAccept((d) -> {
+                final long endTime = System.currentTimeMillis();
+                record(length, startTime, endTime);
+            });
+        }
+        return retVal;
+    }
 }
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -127,11 +127,12 @@ class PerfStats {
 
     public CompletableFuture runAndRecordTime(Supplier<CompletableFuture> fn, long startTime, int length) {
         CompletableFuture  retVal = fn.get();
-        long endTime = System.currentTimeMillis(); 
         if(retVal == null) {
+            final long endTime = System.currentTimeMillis();
             record(length,startTime,  endTime);
         } else {
             retVal = retVal.thenAccept((d) -> {
+                final long endTime = System.currentTimeMillis();
                 record(length,startTime, endTime);
             });
         }
@@ -141,8 +142,8 @@ class PerfStats {
 
     public CompletableFuture writeAndRecordTime(Supplier<CompletableFuture> fn, int length, boolean blocking ) {
         CompletableFuture  retVal=null;
-        long startTime, endTime;
-        startTime = System.currentTimeMillis();
+        final long startTime = System.currentTimeMillis();
+ 
         try {
              retVal = fn.get();
              if (blocking) {
@@ -151,12 +152,13 @@ class PerfStats {
         } catch (Exception e) {
              e.printStackTrace();
         }
-        endTime = System.currentTimeMillis();
         
         if(retVal == null) {
+            final long endTime = System.currentTimeMillis(); 
             record(length, startTime,  endTime);
         } else {
             retVal = retVal.thenAccept((d) -> {
+                final long endTime = System.currentTimeMillis(); 
                 record(length, startTime, endTime);
             });
         }
--- a/src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
+++ b/src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- * <p>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
+++ b/src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
@@ -144,9 +144,8 @@ public class PravegaPerfTest {
 
             factory = new ClientFactoryImpl(scopeName, controller);
 
-
             if (consumerCount > 0 ) {
-               readerGroup = streamHandle.createReaderGroup();  
+               readerGroup = streamHandle.createReaderGroup();
                drainStats = new PerfStats("Draining", reportingInterval, messageSize);
                consumeStats = new PerfStats("Reading", reportingInterval, messageSize);
                ReaderWorker.totalEvents = new AtomicInteger(consumerCount * eventsPerSec * runtimeSec);
@@ -336,17 +335,17 @@ public class PravegaPerfTest {
 }
 
 class StreamHandler{
-      String scope;
-      String stream;
-      String controllerUri;   
+      final String scope;
+      final String stream;
+      final String controllerUri;   
       ControllerImpl controller; 
       StreamManager streamManager;
       StreamConfiguration streamconfig;
       ReaderGroupManager readerGroupManager; 
       ReaderGroup readerGroup;
       ScheduledExecutorService bgexecutor;
-      int segCount;
-      int timeout;
+      final int segCount;
+      final int timeout;
 
 
       StreamHandler(String scope, String stream, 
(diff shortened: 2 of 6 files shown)


src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -118,44 +100,62 @@ public synchronized void printAll() {
...
-            int index = (int) (percentiles[i] * size);
+            int index = (int) (percentiles[i] * count);
             values[i] = latencies[index];
         }
         return values;
     }
 
-    public CompletableFuture runAndRecordTime(Supplier<CompletableFuture> fn, long startTime, int length, Executor executor) {
-        int iter = this.iteration++;
+    public CompletableFuture runAndRecordTime(Supplier<CompletableFuture> fn, long startTime, int length) throws Exception {
```
tkaitchuck: This method takes a Supplier and then immediately call get on it and assigns it to retVal.
This is needlessly complicated. It could just take retVal. There would be no difference.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
                     // event ingestion
-                    long now = System.currentTimeMillis();
-                    retFuture = produceStats.runAndRecordTime(() -> {
-                                String key;
-                                if (isRandomKey) {
-                                    key = Integer.toString(producerId + new Random().nextInt());
-                                } else {
-                                    key = Integer.toString(producerId);
-                                }
+                    retFuture = stats.writeAndRecordTime(() -> {
```
tkaitchuck: I think this code would be simpler with a Timer. See the class Timer in Pravega common.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
 
-    /**
-     * A Sensor reader class that reads the temperative data
-     */
-    private static class ReaderWorker implements Runnable {
-       @Setter
-        public static AtomicInteger totalEvents;
-        private EventStreamReader<String> reader;
-        String readerId;
+class ReaderWorker implements Callable<Void> {
```
tkaitchuck: static?

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
+       public static AtomicInteger totalEvents;
```
tkaitchuck: I think this should be non-static.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
+       private EventStreamReader<String> reader;
```
tkaitchuck: final

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
+       private final int secondsToRun;
+       private final long StartTime;
+       private final int timeout;
+       private final PerfStats stats;
+       String readerId;
```
tkaitchuck: final

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
+            this.timeout = timeout;
+
+            reader = factory.createReader(
+                    this.readerId, readergrp, new JavaSerializer<String>(), ReaderConfig.builder().build());
+       }
+
+       void cleanupEvents(PerfStats drainStats) throws Exception {
+            EventRead<String> result;
+            //System.out.format("******** Draining events from %s/%s%n", scopeName, streamName);
+            do {
```
tkaitchuck: There is no point in a do/while if the condition is "true". Perhaps it should be made conditional on result's event not being null?

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
+                long startTime = System.currentTimeMillis();
+                result = reader.readNextEvent(timeout);
+                if(result.getEvent()!=null) {
+                    drainStats.runAndRecordTime(() -> {
+                    return null;
```
tkaitchuck: indentation

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
+       }
+
+       void cleanupEvents(PerfStats drainStats) throws Exception {
+            EventRead<String> result;
+            //System.out.format("******** Draining events from %s/%s%n", scopeName, streamName);
+            do {
+                long startTime = System.currentTimeMillis();
+                result = reader.readNextEvent(timeout);
+                if(result.getEvent()!=null) {
+                    drainStats.runAndRecordTime(() -> {
```
tkaitchuck: why do we need a lambda function that just returns null?

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
+            this.StartTime=start;
+            this.stats = stats; 
+            this.timeout = timeout;
+
+            reader = factory.createReader(
+                    this.readerId, readergrp, new JavaSerializer<String>(), ReaderConfig.builder().build());
+       }
+
+       void cleanupEvents(PerfStats drainStats) throws Exception {
+            EventRead<String> result;
```
tkaitchuck: It's strange to have a variable named result that is not returned.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -400,76 +602,70 @@ public void run() {
...
+            this.secondsToRun = secondsToRun; 
+            this.StartTime=start;
+            this.stats = stats; 
+            this.timeout = timeout;
+
+            reader = factory.createReader(
+                    this.readerId, readergrp, new JavaSerializer<String>(), ReaderConfig.builder().build());
+       }
+
+       void cleanupEvents(PerfStats drainStats) throws Exception {
```
tkaitchuck: Private. Also Narrow this exception type.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
tkaitchuck: Probably private too right?

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
             runLoop(sendFunction());
-            latch.countDown();
-        }
-    }
+            return null;   
+      }
+}
+
 
+class TransactionWriterWorker extends WriterWorker {
```
tkaitchuck: static

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+      private Transaction<String> transaction;
+      private final int transactionsPerCommit;
+      private int eventCount = 0;
```
tkaitchuck: This value is incremented, but never read.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
 
-    private static class WriterWorker implements Runnable {
+    private static class StartLocalService {
+        static final int PORT = 9090;
+        static final String SCOPE = "Scope";
+        static final String STREAM_NAME = "aaj";
+    }
+}
+
+class StreamHandler{
```
tkaitchuck: This should probably be a top level class.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+      String scope;
```
tkaitchuck: These should all be final

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+               this.readerGroup = null;
+               this.bgexecutor = bgexecutor;
+               streamManager = StreamManager.create(new URI(uri));
+               streamManager.createScope(scope);
+               streamconfig = null;
+
+       } 
+
+
+       boolean create() throws Exception {
```
tkaitchuck: Only declare the exception types your code actually throws. Here and below.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+                                                 .scalingPolicy(ScalingPolicy.fixed(segCount))
+                                                 .build();
+   
+               return streamManager.createStream(scope, stream,streamconfig);
+       }  
+
+       boolean scale() throws Exception {
+               StreamSegments segments = controller.getCurrentSegments(scope, stream).join();
+               final int nseg = segments.getSegments().size();
+               System.out.println("Current segments of the stream: "+stream+ " = " + nseg);
```
tkaitchuck: Use a logger. 
Lombok can help

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+               System.out.println("The stream: " + stream + " will be manually scaling to "+ segCount+ " segments");
+
+               /*
+                * Note that the Upgrade stream API does not change the number of segments;
+                * but it indicates with new number of segments.
+                * after calling update stream , manual scaling is required
+                */
+               if (!streamManager.updateStream(scope, stream, streamconfig)) {
+                    System.out.println("Could not able to update the stream: " + stream + " try with another stream Name");
+                    return false;
```
tkaitchuck: Boolean error codes is strongly discouraged in Java. 
This should be an exception of an appropriate type with the string from the line above.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+                                                               Collectors
+                                                              .toMap(x -> x * keyRangeChunk,
+                                                                      x->(x + 1) * keyRangeChunk));
+               final List<Long> segmentList = segments.getSegments()
+                                                      .stream()
+                                                      .map(Segment::getSegmentId)
+                                                      .collect(Collectors.toList());
+
+               /*
+                 System.out.println("The key ranges are");
```
tkaitchuck: Don't commit commented out code.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+                                             .get().getSegments().size());
+               return true;
+
+       } 
+
+       boolean recreate() throws Exception {
+               System.out.println("Sealing and Deleteing the stream : " + stream + " and then recreating the same");
+               CompletableFuture<Boolean> sealStatus =  controller.sealStream(scope, stream);
+               if (!sealStatus.get(timeout, TimeUnit.SECONDS)) {
+                    System.out.println("ERROR : Segment sealing operation on stream "+ stream + " did not complete");
```
tkaitchuck: again error handling.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
 
-        final EventStreamWriter<String> producer;
-        private final int producerId;
-        private final int eventsPerSec;
-        private final int secondsToRun;
-        private final boolean isTransaction;
 
-        WriterWorker(int sensorId, int eventsPerSec, int secondsToRun, boolean isTransaction, boolean isRandomKey,
-                     ClientFactory factory) {
+class WriterWorker implements Callable<Void> {
```
tkaitchuck: Should this be a top level class?

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+            // producer.close();
+ 
+            //Wait for the last packet to get acked
+            retFuture.get();
         }
 
-        @Override
-        public void run() {
+      @Override
+      public Void call() throws Exception  {
```
tkaitchuck: Just because an interface allows Exception don't mean you can't throw a subclass or none at all. 
Please narrow this to what is actually thrown.

src/main/java/com/emc/pravega/perf/PerfStats.java
kmgowda: here , the sendFunction is override in WriterWorker and TransactionWriterWorker ; hence the supplier is required.  the writeAndRecordTime is invoked by both WriterWorker and TransactionWriterWorker with different sendFunction implementation. 
so, supplier required here.   even though it may look complicated, but supplier is anyway required to pass the lambda function to another method.. please let me know if i am missing anything here

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: well for this i request i can take it in another PR please.
because , somewhere i read the System.nanoTime() is not rhread safe. need to check other alternative.
any change in this would affect the calculations done in the code.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes, total Events is static variable accessed my multiple readers

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: sorry i dint get this.. "call" method can throw an exception which propagated to main. hence changed it from Runnable to callable and suitable of invokeAll also; so that no need of counterlatch. i have made it simple i  believe

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes , I have done this

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes , I have done this

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes . done

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: because runAnd Recordtime check if its null ; then calls record method to store the time stamps

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: i want to keep this generic type, are you see any problem with this?

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: its not ;

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: its used in this comparison
 if (eventCount >= transactionsPerCommit)

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: do i need to move above the main?

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes done..for other similar variables too

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: tom , please  i will do this in later PR.. I just retained the existing system/style of reporting for now;

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: do mean i need to move this above main

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: i have renamed this variable to "event" for better readabliity

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
-        /**
-         * Executes the given method over the producer with configured settings.
-         * @param fn The function to execute.
-         */
-        void runLoop(BiFunction<String, String, CompletableFuture> fn) {
+      /**
+        * Executes the given method over the producer with configured settings.
+        * @param fn The function to execute.
+        */
+      void runLoop(BiFunction<String, String, CompletableFuture> fn) throws Exception {
```
tkaitchuck: Please throw the specific type of exception that you need.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
tkaitchuck: sendFunction() does not throw. runLoop throws, but it is throwing Exception and can be narrowed to a more speffic type. (see my comment there). Once that is done. this function can similarly be narrowed.
Because the interface declares that you can throw Exception, it is OK for your implementation to throw just a subtype for instance IOException. 
This makes the code more readable, and may mean clearer exception handling code in the caller.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
tkaitchuck: Please move this class into it's own file

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
tkaitchuck: Please move this into it's own file.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
tkaitchuck: This is not generic. If you want to parameter your exception go ahead. This is forcing a single catch block on the caller where they don't know how to handle things in a differentiated way. If your only intention is to signal an unrecoverable problem, then you can throw RuntimeException and you don't need this on the signature at all. If you want to signal a specific handle-able problem then declare the type. 
See our guide:
https://github.com/pravega/pravega/wiki/Contributing#exceptions

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
tkaitchuck: Did you push the change?

src/main/java/com/emc/pravega/perf/PerfStats.java
kmgowda: hi Tom, 
    yes I got your point, but this function runAndRecordTime similar to writeAndRecordTime
which actually takes the Supplier , invocation is the writeevent , its similar to that; i am trying to maintain the same structure here.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes,not in this PR please, because lot of places there are System.currentTimeMillis is used and everything i have to change and latency , throughput calculation also chnages;  so , i request i will do this another PR

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes , its done

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: no, this is AtomicInteger  static used by multiple  readers

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes, i have taken care this now.. all the methods will throw the specific errors

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes, its done

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes, I have done this now

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes , done

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: thanks Tom

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes, its done now

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes, i have done this .. i defined my own checked Exceptions functional interface

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -31,81 +32,62 @@
     private long windowStartTime;
     private long start;
     private long windowStart;
-    private long[] latencies;
-    private int sampling;
+    private ArrayList<Double> latencies;
```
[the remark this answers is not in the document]
kmgowda: i have made some of the variables as final.
GurdedBy requires some more analysis ;  which may reorg the code. I will take it as another PR please.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -63,116 +79,161 @@
...
     private static int messageSize = 100;
     private static String streamName = StartLocalService.STREAM_NAME;
-    private static ClientFactory factory = null;
-    private static boolean onlyWrite = true;
-    private static boolean blocking = false;
-    // How many producers should we run concurrently
-    private static int producerCount = 20;
-    private static int consumerCount = 20;
-    // How many events each producer has to produce per seconds
+    private static String scopeName = StartLocalService.SCOPE;
```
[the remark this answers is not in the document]
kmgowda: those are default values , depending on the user input values changes, so i have to use them as it is

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
                     runKafka = Boolean.parseBoolean(commandline.getOptionValue("kafka"));
                 }
 
+                if (commandline.hasOption("segments")) {
```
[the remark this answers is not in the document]
kmgowda: sure, full lombok changes should be another PR.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -63,116 +79,161 @@
...
+                              .collect(Collectors.toList());
+              }       
+
+           } else {
+             writers = null;
+             produceStats = null;
+           } 
+
+
+           final List<Callable<Void>> workers = new ArrayList() {{
```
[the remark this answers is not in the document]
kmgowda: yes, I have done the changes with Steam , should be OK now.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -63,116 +79,161 @@
...
+              produceStats.printTotal(endTime);
+           }
+
+           if ( consumeStats != null ) {
+              consumeStats.printTotal(endTime);
+
+           }
+
+        } catch (Exception e) {
+            e.printStackTrace();
```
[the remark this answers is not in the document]
kmgowda: yes, its removed now

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -63,116 +79,161 @@
...
+        } finally {
+           fjexecutor.shutdown();
+           fjexecutor.awaitTermination(runtimeSec, TimeUnit.SECONDS);
         }
+
```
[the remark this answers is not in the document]
kmgowda: yes

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -259,131 +314,278 @@ private static void parseCmdLine(String[] args) {
...
+                    segmentCount = producerCount; 
+                }
+
+                if (commandline.hasOption("recreate")) {
+                    recreate = Boolean.parseBoolean(commandline.getOptionValue("recreate"));
+                }
+
             }
         } catch (Exception nfe) {
             System.out.println("Invalid arguments. Starting with default values");
```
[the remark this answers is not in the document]
kmgowda: yes, its taken care now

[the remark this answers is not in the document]
kmgowda: this will throw the ParseException

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
kmgowda: yes, now the Exceptions are no more generic , except in the main.. all the other places , the excpetions are specific now

src/main/java/com/emc/pravega/perf/PerfStats.java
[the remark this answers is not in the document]
tkaitchuck: They are. I checked.

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -118,44 +103,62 @@ public synchronized void printAll() {
...
-                record(iter, (int) (System.currentTimeMillis() - startTime) * 1000, length, System.nanoTime());
+                final long endTime = System.currentTimeMillis();
+                record(length,startTime, endTime);
             });
         }
         return retVal;
 
     }
+
+    public CompletableFuture writeAndRecordTime(SupplierWithCE<CompletableFuture, TxnFailedException> fn, int length) throws TxnFailedException {
```
tkaitchuck: If these two methods are going to be consistent with one another, then this one should take startTime as an argument like runAndRecordTime does.

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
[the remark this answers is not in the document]
tkaitchuck: I don't see this change

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
```diff
@@ -0,0 +1,97 @@
...
+
+
+public class PravegaReaderWorker implements Callable<Void> {
+    public static AtomicInteger totalEvents;
+    private final EventStreamReader<String> reader;
+    private final int secondsToRun;
+    private final long StartTime;
+    private final int timeout;
+    private final PerfStats stats;
+    String readerId;
```
tkaitchuck: final

src/main/java/com/emc/pravega/perf/PravegaStreamHandler.java
```diff
@@ -0,0 +1,162 @@
...
+import io.pravega.client.stream.impl.StreamImpl;
+import io.pravega.client.ClientConfig;
+import io.pravega.client.stream.Stream;
+
+
+public class PravegaStreamHandler {
+    final String scope;
+    final String stream;
+    final String controllerUri;
+    ControllerImpl controller;
```
tkaitchuck: these can be final too

src/main/java/com/emc/pravega/perf/PravegaStreamHandler.java
```diff
@@ -0,0 +1,162 @@
...
+                .collect(Collectors.toList());
+
+        CompletableFuture<Boolean> scaleStatus = controller.scaleStream(new StreamImpl(scope,stream),
+                segmentList,
+                keyRanges,
+                bgexecutor).getFuture();
+
+
+        if (!scaleStatus.get(timeout, TimeUnit.SECONDS)){
+            throw new InterruptedException("ERROR : Scale operation on stream "+ stream+" did not complete");
```
tkaitchuck: You should use timeoutException. Interrupted refers to a Thread.interupt which is different. (anywhere else this pattern applies too)

src/main/java/com/emc/pravega/perf/PerfStats.java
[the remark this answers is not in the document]
kmgowda: thanks

src/main/java/com/emc/pravega/perf/PerfStats.java
kmgowda: no, its need not be 100% consistent to each other; earlier runAndRecordTime time was used for write also, I was not comfortable  with timestamping locations in the code, hence i wrote new method:writeAndRecordTime
in the next PR i will try to merge bot the methods.

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- * 
+ * <p>
```
RaulGracia: Please, remove this: I know that is how IntelliJ automatically formats the header, but this is not how the header is in all other files in Pravega.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.concurrent.ExecutionException;
+import io.pravega.client.stream.EventStreamReader;
+import io.pravega.client.ClientFactory;
+import io.pravega.client.stream.impl.JavaSerializer;
+import io.pravega.client.stream.ReaderConfig;
+import io.pravega.client.stream.EventRead;
+import io.pravega.client.stream.ReinitializationRequiredException;
+
+public class PravegaReaderWorker implements Callable<Void> {
+    public static AtomicInteger totalEvents;
```
tkaitchuck: not static, final.

src/main/java/com/emc/pravega/perf/PravegaStreamHandler.java
```diff
@@ -0,0 +1,159 @@
...
+            throw new TimeoutException("ERROR : stream: " + stream + " recreation failed");
+        }
+    }
+
+    ReaderGroup createReaderGroup() throws URISyntaxException {
+        if (readerGroup != null) {
+            return readerGroup;
+        }
+
+        readerGroupManager = ReaderGroupManager.withScope(scope,
```
tkaitchuck: This value is only used in this method, it can be made into a local variable.

src/main/java/com/emc/pravega/perf/PravegaStreamHandler.java
```diff
@@ -0,0 +1,159 @@
...
+            throw new TimeoutException("ERROR : stream: " + stream + " delete failed");
+        }
+
+        if (!streamManager.createStream(scope, stream, streamconfig)) {
+            throw new TimeoutException("ERROR : stream: " + stream + " recreation failed");
+        }
+    }
+
+    ReaderGroup createReaderGroup() throws URISyntaxException {
+        if (readerGroup != null) {
```
tkaitchuck: This is not thread safe.

tkaitchuck: Additionally I'm not sure why it needs to guard against it being called multiple times. I only see one call site, and it only is invoked once in the lifetime of the process. I think that readerGroup can simply be returned and not stored as a member variable.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
```diff
@@ -0,0 +1,89 @@
...
```
tkaitchuck: final

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -0,0 +1,55 @@
...
+
+package com.emc.pravega.perf;
+
+import java.util.concurrent.CompletableFuture;
+import io.pravega.client.ClientFactory;
+import io.pravega.client.stream.Transaction;
+import io.pravega.client.stream.TxnFailedException;
+
+public class PravegaTransactionWriterWorker extends PravegaWriterWorker {
+    private Transaction<String> transaction;
```
tkaitchuck: `transaction` and `eventCount` below are accessed in the writeData method which is called from a forkJoin pool and hence may have call coming from multiple threads. But their assignments are not guarded by any lock, so this is a race condition. 
Specifically below, I can see this resulting in a potential NPE, a potential use of a transaction after commit, and multiple ways to miscount of events.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
kmgowda: yes, its done now

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: ok, now evenCount is Atomic integer and transaction is the Atomic reference for the producer.begintransaction ; should be OK now.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -0,0 +1,57 @@
...
+            messageSize, start, factory, stats, streamName);
+
+        this.transactionsPerCommit = transactionsPerCommit;
+        eventCount = new AtomicInteger(0);
+        transaction = new AtomicReference<Transaction<String>>(producer.beginTxn());
+    }
+
+    @Override
+    public CompletableFuture writeData(String key, String data) throws TxnFailedException {
+        transaction.get().writeEvent(key, data);
```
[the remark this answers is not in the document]
kmgowda: yes , done

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -0,0 +1,57 @@
...
+        if (eventCount.incrementAndGet() >= transactionsPerCommit) {
+            eventCount.set(0);
+            transaction.get().commit();
+            transaction.set(producer.beginTxn());
```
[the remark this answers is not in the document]
kmgowda: thanks.. yes its done

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
```diff
@@ -0,0 +1,89 @@
...
+                        String readergrp, int timeout, long totalEvents) {
+        this.readerId = Integer.toString(readerId);
+        this.secondsToRun = secondsToRun;
+        this.StartTime = start;
+        this.stats = stats;
+        this.timeout = timeout;
+        this.totalEvents = totalEvents;
+
+        reader = factory.createReader(
+            this.readerId, readergrp, new JavaSerializer<String>(), ReaderConfig.builder().build());
```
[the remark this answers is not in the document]
kmgowda: yes , done

src/main/java/com/emc/pravega/perf/PravegaWriterWorker.java
```diff
@@ -0,0 +1,110 @@
...
+                        ClientFactory factory, PerfStats stats, String streamName) {
+        this.producerId = sensorId;
+        this.eventsPerSec = eventsPerSec;
+        this.secondsToRun = secondsToRun;
+        this.StartTime = start;
+        this.stats = stats;
+        this.isRandomKey = isRandomKey;
+        this.messageSize = messageSize;
+        this.producer = factory.createEventWriter(streamName,
+            new JavaSerializer<String>(),
```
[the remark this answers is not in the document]
kmgowda: yes , done

kmgowda:
thanks Tom

PR #19, opened 2019-01-10
Refactor pravega writer and readers with the abstract Writers and readers

kmgowda:
The new abstract classes WriteWorker and ReadWorkers are created.
These classes does the performance bench-marking whereas the
pravega writer and pravega readers classes does the data writes and reads.

### DIFF
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- * <p>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -26,7 +26,10 @@ import java.time.Duration;
 import io.pravega.client.stream.TxnFailedException;
 import io.pravega.client.stream.ReinitializationRequiredException;
 
-class PerfStats {
+/**
+ *  class for Performance statistics.
+ */
+public class PerfStats {
     final private int messageSize;
     final private String action;
     private Instant windowStartTime;
@@ -57,7 +60,7 @@ class PerfStats {
         this.messageSize = messageSize;
     }
 
-    public synchronized void record(int bytes, Instant startTime, Instant endTime) {
+    private synchronized void record(int bytes, Instant startTime, Instant endTime) {
         this.iteration++;
         this.windowBytes += bytes;
         this.windowCount++;
@@ -100,6 +103,11 @@ class PerfStats {
         */
     }
 
+    /**
+     * print the final performance statistics.
+     *
+     * @param endTime        endtime to performance benchmarking.
+     */
     public synchronized void printTotal(Instant endTime) {
         final long elapsed = Duration.between(start, endTime).toMillis();
         double recsPerSec = 1000.0 * iteration / (double) elapsed;
@@ -125,6 +133,14 @@ class PerfStats {
         return values;
     }
 
+    /**
+     * record the data write/read time of given length of data.
+     *
+     * @param retVal         future to wait for.
+     * @param startTime      starting time
+     * @param length         length of data read/written
+     * @return a completable future for recording the end time.
+     */
     public CompletableFuture recordTime(CompletableFuture retVal, Instant startTime, int length) {
         if (retVal == null) {
             final Instant endTime = Instant.now();
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- * <p>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -17,95 +17,75 @@
  */
 package com.emc.pravega.perf;
 
-
-
-
 import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
-import java.util.function.Supplier;
+import io.pravega.client.stream.TxnFailedException;
+import io.pravega.client.stream.ReinitializationRequiredException;
 
 class PerfStats {
-    private int messageSize;
-    private String action;
+    final private int messageSize;
+    final private String action;
     private long windowStartTime;
     private long start;
     private long windowStart;
-    private long[] latencies;
-    private int sampling;
+    private ArrayList<Double> latencies;
     private int iteration;
-    private int index;
     private long count;
     private long bytes;
-    private int maxLatency;
-    private long totalLatency;
+    private double maxLatency;
+    private double totalLatency;
     private long windowCount;
-    private int windowMaxLatency;
-    private long windowTotalLatency;
     private long windowBytes;
-    private long reportingInterval;
+    final private long reportingInterval;
 
-    public PerfStats(String action, long numRecords, int reportingInterval, int messageSize) {
-        if ( numRecords != 0 ) {
-            this.action = action;
-            this.start = System.currentTimeMillis();
-            this.windowStartTime = System.currentTimeMillis();
-            this.windowStart = 0;
-            this.index = 0;
-            this.iteration = 0;
-            this.sampling = (int) (numRecords / Math.min(numRecords, 500000));
-            this.latencies = new long[(int) (numRecords / this.sampling)];
-            this.index = 0;
-            this.maxLatency = 0;
-            this.totalLatency = 0;
-            this.windowCount = 0;
-            this.windowMaxLatency = 0;
-            this.windowTotalLatency = 0;
-            this.windowBytes = 0;
-            this.totalLatency = 0;
-            this.reportingInterval = reportingInterval;
-            this.messageSize = messageSize;
-        }
+    public PerfStats(String action, int reportingInterval, int messageSize) {
+        this.action = action;
+        this.start = System.currentTimeMillis();
+        this.windowStartTime = System.currentTimeMillis();
+        this.windowStart = 0;
+        this.iteration = 0;
+        this.latencies = new ArrayList<Double>();
+        this.maxLatency = 0;
+        this.totalLatency = 0;
+        this.windowCount = 0;
+        this.windowBytes = 0;
+        this.reportingInterval = reportingInterval;
+        this.messageSize = messageSize;
     }
 
-    public synchronized void record(int iter, int latencyMicro, int bytes, long time) {
-        this.count++;
-        this.bytes += bytes;
-        this.totalLatency += latencyMicro;
-        this.maxLatency = Math.max(this.maxLatency, latencyMicro);
-        this.windowCount++;
+    public synchronized void record(int bytes, long startTime, long endTime) {
+        this.iteration++;
         this.windowBytes += bytes;
-        this.windowTotalLatency += latencyMicro;
-        this.windowMaxLatency = Math.max(windowMaxLatency, latencyMicro);
-        if (iter % this.sampling == 0) {
-            this.latencies[index] = latencyMicro;
-            this.index++;
-        }
-        /* maybe report the recent perf */
-        if (count - windowStart >= reportingInterval) {
-            printWindow();
+        this.windowCount++;
+        /* did we arrived at reporting time */
+        if ((endTime - windowStartTime) >= reportingInterval) {
+            printWindow(endTime);
             newWindow(count);
         }
     }
 
-    private void printWindow() {
-        long elapsed = System.currentTimeMillis() - windowStartTime;
+    private void printWindow(long endTime) {
+        long elapsed = endTime - windowStartTime;
+        double latency = (double) (elapsed / (double) windowCount);
         double recsPerSec = 1000.0 * windowCount / (double) elapsed;
         double mbPerSec = 1000.0 * this.windowBytes / (double) elapsed / (1024.0 * 1024.0);
-        System.out.printf("%d records %s, %.1f records/sec (%.5f MB/sec), %.1f ms avg latency, %.1f max latency.\n",
-                windowCount, action, recsPerSec, mbPerSec, windowTotalLatency / ((double) windowCount * 1000.0),
-                (double) windowMaxLatency / 1000.0);
-        System.out.printf(" WINDOW: %d, %d, %.1f ,%.5f MB/sec, %.1f, %.1f \n",
-                messageSize, windowCount, recsPerSec, mbPerSec, windowTotalLatency / ((double) windowCount * 1000.0),
-                (double) windowMaxLatency / 1000.0);
+
+        this.bytes += this.windowBytes;
+        this.totalLatency += latency;
+        this.maxLatency = Math.max(this.maxLatency, latency);
+        this.latencies.add(latency);
+        this.count++;
+
+        System.out.printf("%8d records %s, %9.1f records/sec, %9.3f MB/sec, %7.4f ms avg latency.\n",
+            windowCount, action, recsPerSec, mbPerSec, latency);
     }
 
     private void newWindow(long currentNumber) {
         this.windowStart = currentNumber;
         this.windowStartTime = System.currentTimeMillis();
         this.windowCount = 0;
-        this.windowMaxLatency = 0;
-        this.windowTotalLatency = 0;
         this.windowBytes = 0;
     }
 
@@ -118,44 +98,41 @@ class PerfStats {
         */
     }
 
-    public synchronized void printTotal() {
-        long elapsed = System.currentTimeMillis() - start;
-        double recsPerSec = 1000.0 * count / (double) elapsed;
+    public synchronized void printTotal(long endTime) {
+        long elapsed = endTime - start;
+        double recsPerSec = 1000.0 * iteration / (double) elapsed;
         double mbPerSec = 1000.0 * this.bytes / (double) elapsed / (1024.0 * 1024.0);
-        long[] percs = percentiles(this.latencies, 0.5, 0.95, 0.99, 0.999);
+        //double[] percs = percentiles(this.latencies, 0.5, 0.95, 0.99, 0.999);
         System.out.printf(
-                "%d records sent, %f records/sec (%.5f MB/sec), %.2f ms avg latency, %.2f ms max " + "latency, %.2f " +
-                        "ms 50th, %.2f ms 95th, %.2f ms 99th, %.2f ms 99.9th.\n",
-                count, recsPerSec, mbPerSec, totalLatency / ((double) count * 1000.0), (double) maxLatency / 1000.0,
-                percs[0] / 1000.0, percs[1] / 1000.0, percs[2] / 1000.0, percs[3] / 1000.0);
-        System.out.printf(
-                " %s FINAL:, %d, %.5f MB/sec, %.2f, %.2f, %.2f, %.2f, %.2f, %.2f\n", this.action,
-                messageSize, mbPerSec, totalLatency / ((double) count * 1000.0), (double) maxLatency / 1000.0,
-                percs[0] / 1000.0, percs[1] / 1000.0, percs[2] / 1000.0, percs[3] / 1000.0);
+            "%d records %s, %.3f records/sec, %d bytes record size, %.3f MB/sec, %.4f ms avg latency, %.4f ms max latency\n",
+            iteration, action, recsPerSec, messageSize, mbPerSec, totalLatency / ((double) count), (double) maxLatency);
+        /*  
+        System.out.printf("latencies percentiles:  %.4f ms 50th, %.4f ms 95th, %.4f ms 99th, %.4f ms 99.9th.\n",
+                           percs[0], percs[1], percs[2], percs[3]);
+        */
+
     }
 
-    private long[] percentiles(long[] latencies, double... percentiles) {
-        long size = Math.min(count, latencies.length);
-        Arrays.sort(latencies, 0, (int) size);
-        long[] values = new long[percentiles.length];
+    private double[] percentiles(double[] latencies, double... percentiles) {
+        Arrays.sort(latencies, 0, (int) count);
+        double[] values = new double[percentiles.length];
         for (int i = 0; i < percentiles.length; i++) {
-            int index = (int) (percentiles[i] * size);
+            int index = (int) (percentiles[i] * count);
             values[i] = latencies[index];
         }
         return values;
     }
 
-    public CompletableFuture runAndRecordTime(Supplier<CompletableFuture> fn, long startTime, int length, Executor executor) {
-        int iter = this.iteration++;
-        CompletableFuture  retVal = fn.get();
-        if(retVal == null) {
-            record(iter, (int) (System.currentTimeMillis() - startTime) * 1000, length, System.nanoTime());
+    public CompletableFuture recordTime(CompletableFuture retVal, long startTime, int length) {
+        if (retVal == null) {
+            final long endTime = System.currentTimeMillis();
+            record(length, startTime, endTime);
         } else {
             retVal = retVal.thenAccept((d) -> {
-                record(iter, (int) (System.currentTimeMillis() - startTime) * 1000, length, System.nanoTime());
+                final long endTime = System.currentTimeMillis();
+                record(length, startTime, endTime);
             });
         }
         return retVal;
-
     }
 }
--- a/src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
+++ b/src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
@@ -129,7 +129,6 @@ public class PravegaPerfTest {
 
             if (consumerCount > 0) {
                 readerGroup = streamHandle.createReaderGroup();
-                drainStats = new PerfStats("Draining", reportingInterval, messageSize);
                 consumeStats = new PerfStats("Reading", reportingInterval, messageSize);
 
                 readers = IntStream.range(0, consumerCount)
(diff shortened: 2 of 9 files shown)


src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
```diff
@@ -30,60 +28,44 @@
...
-                if (ret != null) {
-                    stats.recordTime(null, Instant.ofEpochMilli(Long.parseLong(ret.split(",")[0])), ret.length());
-                }
-            } while ((Duration.between(StartTime, Instant.now()).getSeconds() < secondsToRun) &&
-                    (eventCount.incrementAndGet() < totalEvents));
-        } finally {
-            reader.close();
+            return reader.readNextEvent(timeout).getEvent();
+        } catch(ReinitializationRequiredException e){
+            throw new Exception("ReinitializationRequiredException exception while reading data ");
```
tkaitchuck: This doesn't seem useful.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -34,30 +34,33 @@
...
+                messageSize, start, stats, streamName, totalEvents, factory);
 
         this.transactionsPerCommit = transactionsPerCommit;
         eventCount = new AtomicInteger(0);
         transaction = new AtomicReference<Transaction<String>>(producer.beginTxn());
     }
 
     @Override
-    public CompletableFuture writeData(String key, String data) throws TxnFailedException, IllegalStateException {
+    public CompletableFuture writeData(String key, String data) throws IllegalStateException {
```
tkaitchuck: IllegalStateException is unchecked.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -34,30 +34,33 @@
...
-            eventCount.set(0);
-            curTrans.commit();
-            if (!transaction.compareAndSet(curTrans, producer.beginTxn())) {
-                throw new IllegalStateException("WriteData called on the same PravegaTransactionWriterWorker from two threads in parallel.");
+        try {
+            curTrans.writeEvent(key, data);
+            if (eventCount.incrementAndGet() >= transactionsPerCommit) {
+                eventCount.set(0);
+                curTrans.commit();
+                if (!transaction.compareAndSet(curTrans, producer.beginTxn())) {
```
tkaitchuck: This logic is not thread safe.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -34,30 +34,33 @@
...
+                    throw new IllegalStateException("WriteData called on the same PravegaTransactionWriterWorker from two threads in parallel.");
+                }
             }
+        } catch (TxnFailedException e) {
+            throw new IllegalStateException("Write data failed with transaction failure.");
```
tkaitchuck: I'm not sure this qualfies as an illegal state.

src/main/java/com/emc/pravega/perf/PravegaWriterWorker.java
```diff
@@ -32,82 +29,41 @@
...
-
+    public void flush() {
         producer.flush();
-        // producer.close();
+    }
 
-        //Wait for the last packet to get acked
-        retFuture.get();
+    @Override
+    public long eventCountIncrementAndGet() {
```
tkaitchuck: Why is this a public method?

src/main/java/com/emc/pravega/perf/ReaderWorker.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public abstract String readData() throws Exception;
+
+    public abstract void close();
+
+    public abstract long eventCountIncrementAndGet();
+
+    public abstract long eventCountGet();
+
+    public void cleanupEvents(PerfStats drainStats) throws Exception {
```
tkaitchuck: All of these public methods and those elsewhere need to have Javadocs. 
For example I have no idea why "cleanupEvents" would involve invoking "readData".

src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -0,0 +1,87 @@
...
+    public abstract CompletableFuture writeData(String key, String data);
+
+    public abstract void flush();
+
+    public abstract long eventCountIncrementAndGet();
+
+    public abstract long eventCountGet();
+
+    @Override
+    public Void call() throws InterruptedException, ExecutionException, IllegalStateException {
```
tkaitchuck: Here and elsewhere: You don't need to declare runtime exception in the signature.

src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -0,0 +1,87 @@
...
+            final Instant loopStartTime = Instant.now();
+            for (int i = 0; (i < eventsPerSec) && (eventCountIncrementAndGet() <= totalEvents) &&
+                    (Duration.between(loopStartTime, Instant.now()).getSeconds() < 1); i++) {
+
+                // Construct event payload
+                String val = System.currentTimeMillis() + ", " + workerID + ", " + (int) (Math.random() * 200);
+                String payload = String.format("%-" + messageSize + "s", val);
+                String key;
+                if (isRandomKey) {
+                    key = Integer.toString(workerID + new Random().nextInt());
```
tkaitchuck: Don't use random as a one off like this. It is not random. You need to same it as a variable outside the loop so you can use it over and over. 
Also spotbugs should have caught this. Is it not running on this repo? If so we should enable it.

src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -0,0 +1,87 @@
...
+
+                final Instant startTime = Instant.now();
+                retFuture = writeData(key, payload);
+                // event ingestion
+                retFuture = stats.recordTime(retFuture, startTime, payload.length());
+            }
+
+            long timeSpent = Duration.between(loopStartTime, Instant.now()).toMillis();
+            if (timeSpent < 1000) {
+                Thread.sleep(1000 - timeSpent);
```
tkaitchuck: Please file an issue on dealing with over sleeps.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
```diff
@@ -30,60 +28,44 @@
 import io.pravega.client.stream.EventRead;
 import io.pravega.client.stream.ReinitializationRequiredException;
 
-public class PravegaReaderWorker implements Callable<Void> {
+public class PravegaReaderWorker extends ReaderWorker  {
     public final static AtomicInteger eventCount = new AtomicInteger(0);
```
tkaitchuck: I don't think this should be public or static.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
```diff
@@ -30,60 +28,44 @@
...
-        return null;
     }
+
+    @Override
+    public void close(){
+        reader.close();
+    }
+
+    @Override
+    public long eventCountIncrementAndGet() {
```
tkaitchuck: This should not be public

src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -0,0 +1,87 @@
...
+        super(sensorId, eventsPerSec, secondsToRun,
+                isRandomKey, messageSize, start,
+                stats, streamName, totalEvents, 0);
+    }
+
+    public abstract CompletableFuture writeData(String key, String data);
+
+    public abstract void flush();
+
+    public abstract long eventCountIncrementAndGet();
```
tkaitchuck: This should not be public

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -34,30 +34,33 @@
...
+                                   PerfStats stats, String streamName,
+                                   long totalEvents, ClientFactory factory,
                                    int transactionsPerCommit) {
 
         super(sensorId, eventsPerSec, secondsToRun, isRandomKey,
-                messageSize, start, factory, stats, streamName, totalEvents);
+                messageSize, start, stats, streamName, totalEvents, factory);
 
         this.transactionsPerCommit = transactionsPerCommit;
         eventCount = new AtomicInteger(0);
```
tkaitchuck: This is masking eventCount for the super class

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -34,30 +34,33 @@
     PravegaTransactionWriterWorker(int sensorId, int eventsPerSec,
```
tkaitchuck: I think this class could just extend WriterWorker directly.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
kmgowda: i need your help here Tom,   this what my design in brief.
WriterWorker  - An abstract class which implements call method the performance benchmarking.
                           This class will not know anything about pravega or any other streaming apis (like kafka)

PravegaWriterWorker  - Pravega specific writer class for writer benchmarking. This class extends  "WriterWorker  " .. 

so, here is the problem 

methods of PravegaWriterWorker class can throw pravega specific exceptions but  call method (for performance bench-marking) or any other abstract methods of the base abstract class WriterWorker  will not know about it.  So, here WriterWorker   should be generic.
hence, ReinitializationRequiredException  e is converted as generic exception.

please let me know if you want better of dealing the above scenario.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: same as above, since the writedata is abstract method of base class WriterWorker  , this class does not know TxnFailedException of pravega.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: eventCount is changed to transEventCount its Atomic interger and it increments rather than decrementing as per your last review comments. please let me know where exactly you are seeing an issue

src/main/java/com/emc/pravega/perf/PravegaWriterWorker.java
kmgowda: this is overriding the abstract method fromt the abstract class WriterWorker

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
kmgowda: thanks.. I have made this one as private now

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: no, this is an abstract methods must be overridden in derived class , should be public.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: thanks a lot Tom, I have changed this to transEventCount

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: no, here the PravegaTransactionWriterWorker need the producer created by PravegaWriterWorker.

src/main/java/com/emc/pravega/perf/ReaderWorker.java
kmgowda: yes, I have added Javdoc tags

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: I will do this as separate PR please.. I need to lot of stuff during random key till then i am just keeping the existing code

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: sure..  but what is the issue here? this seems ok for me.. if the events are pushed before time, then we can wait for next sec to start the writes right?

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
kmgowda: no, this is an abstract methods must be overridden in derived class , need to be public.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
tkaitchuck: It's fine to convert one exception type to another specific type. But here it's just converting to "Exception" which does not serve any purpose because `ReinitializationRequiredException` already is an Exception. So this rethrowing could be deleted and it wouldn't affect the code in any way besides the string "ReinitializationRequiredException exception while reading data" being printed if the caller logs the exception. 
If you want to create some sort of WorkerFailureException then a conversion might make sense.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
tkaitchuck: yes, but IllegalState is unchecked. I does not need to be on this line. You can still throw it, there is just no need to type it here.

src/main/java/com/emc/pravega/perf/WriterWorker.java
tkaitchuck: Yes, add spotbugs as a separate PR. But please move the construction of random up to line 51.

src/main/java/com/emc/pravega/perf/WriterWorker.java
tkaitchuck: sleep is not guarenteed to wake up after the duration specified. It will wake up some time later. So what happens when you ask it to sleep for 300ms and it sleeps for 3s?

src/main/java/com/emc/pravega/perf/WriterWorker.java
tkaitchuck: abstract methods do not have to be public.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
tkaitchuck: Performing transaction on a Producer is a deprecated API. The new API works by creating a transactional producer from the client factory directly.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
```diff
@@ -30,60 +28,48 @@
...
 
-public class PravegaReaderWorker implements Callable<Void> {
-    public final static AtomicInteger eventCount = new AtomicInteger(0);
-    public final long totalEvents;
+
+/**
+ *  class for Pravega reader/consumer.
+ */
+public class PravegaReaderWorker extends ReaderWorker  {
+    private final static AtomicInteger eventCount = new AtomicInteger(0);
```
tkaitchuck: Making this static means that two instances of this test can never be run in the the same VM. That's a really odd restriction to impose, especially without documentation. Wouldn't a target rate an duration be sufficient to control when to exit?

src/main/java/com/emc/pravega/perf/ReaderWorker.java
```diff
@@ -0,0 +1,98 @@
...
+    public Void call() throws ExecutionException, Exception {
+        String ret = null;
+        try {
+            do {
+                ret = readData();
+                if (ret != null) {
+                    stats.recordTime(null, Instant.ofEpochMilli(Long.parseLong(ret.split(",")[0])), ret.length());
+                }
+            } while ((Duration.between(StartTime, Instant.now()).getSeconds() < secondsToRun) &&
+                    (eventCountIncrementAndGet() < totalEvents));
```
tkaitchuck: eventCountIncrementAndGet() is an instance method that is defined above but it is being compared against a totalEvents count. 
If this is meant to be an instance variable, then it can simply be made into a local variable because it is not used in the child class.
If it is meant to be a global, then why is the child forced to implement it? there is only one possible implementation.

src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public abstract long eventCountGet();
+
+    @Override
+    public Void call() throws InterruptedException, ExecutionException, IllegalStateException {
+        CompletableFuture retFuture = null;
+
+        do {
+            final Instant loopStartTime = Instant.now();
+            for (int i = 0; (i < eventsPerSec) && (eventCountIncrementAndGet() <= totalEvents) &&
```
tkaitchuck: Why do the children need to implement this?
eventCountIncrementAndGet() is an instance method that is defined above but it is being compared against a totalEvents count.
If this is meant to be an instance variable, then it can simply be made into a local variable because it is not used in the child class.
If it is meant to be a global, then why is the child forced to implement it? there is only one possible implementation.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
kmgowda: yes, then in this case ReinitializationRequiredException can be converted to illegalstateExcpetion

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: i dint get this.. it throws IllegalStateException  correct?

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: yes, I have moved  below two lines separate synchronized function: 
   
 private synchronized boolean transIncrementAndCompare() {
        if (transEventCount.incrementAndGet() >= transactionsPerCommit) {
            transEventCount.set(0);
            return true;
        }
        return false;
    }

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: yes, I do feel so.. but can you suggest which generic exception i can use? shall i use RuntimeException?

src/main/java/com/emc/pravega/perf/PravegaWriterWorker.java
kmgowda: ok, i have removed them now

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: do want me to remove ExecutionException?

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: thanks.. yes , its done.

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
kmgowda: removed these methods

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: yes, but i will do as part of  : https://github.com/pravega/pravega-benchmark/issues/16

src/main/java/com/emc/pravega/perf/PravegaReaderWorker.java
kmgowda: yes, its removed now

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: removed

src/main/java/com/emc/pravega/perf/ReaderWorker.java
kmgowda: removed

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: yes.. this was bothering me since quite some time.. I followed the kafka benchmarking tool logic, so , i have changed the eventspecsec to events per producer /consumer. this new logic eliminates the total count and threadsleeps.

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: i have removed all except the writedata and event.. these abstract methods need to extended by derived classes; so i have to keep them public.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
tkaitchuck: The method can throw IllegalStateException without it being declared in the function signature.
https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
tkaitchuck: RuntimeException is fine, just be sure to pass `e` as the second arg to the constructor so that the trace and cause is not lost.

src/main/java/com/emc/pravega/perf/WriterWorker.java
tkaitchuck: All three of these do not need to be enumerated here.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
tkaitchuck: This doesn't fix the problem if transactionsPerCommit is 1.

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- *
+ * <p>
```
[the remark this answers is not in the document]
kmgowda: done

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -6,9 +6,7 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- *
```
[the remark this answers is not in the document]
kmgowda: done

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -70,17 +68,16 @@
     private static int producerCount = 0;
     private static int consumerCount = 0;
     private static int segmentCount = 0;
-    private static int eventsPerSec = 40;
-    private static int runtimeSec = 10;
+    private static int eventsPerWorker = 40;
```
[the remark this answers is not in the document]
kmgowda: i have changed this to 3000 by default;
idea here is , if the user has not supplied for how long he wants to execute the benchmarking then it will run for 24hrs or  writes/reads 3000 events by one producer/reader which ever is earlier/less.

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: done

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: since the call() invokes future.get , it requires  InterruptedException, ExecutionException to checked , so i have removed only IllegalStateException

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: yes , its done

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: ok, then for time being for this PR, I have done some this:
        synchronized (this) {
                if (transEventCount.incrementAndGet() >= transactionsPerCommit) {
                    transEventCount.set(0);
                    curTrans.commit();
                    if (!transaction.compareAndSet(curTrans, producer.beginTxn())) {
                        throw new IllegalStateException("WriteData called on the same PravegaTransactionWriterWorker from two threads in parallel.");
                    }
                }

synchronized count and commit.

while addressing https://github.com/pravega/pravega-benchmark/issues/16, I will revisit this logic to minimize the sync block

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: done

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -29,35 +29,39 @@
...
 
-        curTrans.writeEvent(key, data);
-        if (eventCount.incrementAndGet() >= transactionsPerCommit) {
-            eventCount.set(0);
-            curTrans.commit();
-            if (!transaction.compareAndSet(curTrans, producer.beginTxn())) {
-                throw new IllegalStateException("WriteData called on the same PravegaTransactionWriterWorker from two threads in parallel.");
+        try {
+            curTrans.writeEvent(key, data);
+            synchronized (this) {
```
tkaitchuck: This solves the race condition. Now you can simplify transEventCount can be a regular integer guardedBy("this"). And the if block that throws the illegal state can be removed.

Additionally you can also do the same with transaction. It can be made into a normal member guarded by 'this', and you don't need the atomic reference any more.

src/main/java/com/emc/pravega/perf/ReaderWorker.java
```diff
@@ -0,0 +1,83 @@
...
+     * close the consumer/reader.
+     */
+    public abstract void close();
+
+    /**
+     * read all the data from reader/consumer.
+     *
+     * @param drainStats object which collects the details about the data read till there is no data to read from stream
+     */
+    public void cleanupEvents(PerfStats drainStats) {
```
tkaitchuck: It's weird that this is called 'cleanup' it doesn't actually result in anything being deleted. Maybe it should be called "readRemainingEvents" or something.

src/main/java/com/emc/pravega/perf/ReaderWorker.java
```diff
@@ -0,0 +1,83 @@
...
+        } while (ret != null);
+        drainStats.printTotal(Instant.now());
+    }
+
+    @Override
+    public Void call() {
+        String ret = null;
+        try {
+
+            for (int i = 0; (i < eventsPerWorker) &&
```
tkaitchuck: Should readers be capped based on a count per worker? The events won't be uniformly distributed, so if some exit early because they hit the count it's going to mean some events go unread. Perhaps readers should just be a time thing, or perhaps they could have some sort of 'stop()' method. (If you want to do that in a different PR that's fine)

src/main/java/com/emc/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: yes , its done

src/main/java/com/emc/pravega/perf/ReaderWorker.java
kmgowda: this cleanup gets invoked if both producer and consumers are specified with bench marking.. before the producer starts writing , this methods will clean the existing events. since its invoked before reading its kept as cleanupEvents and i am just maintain this method as it is.  i will revisit this soon.

src/main/java/com/emc/pravega/perf/ReaderWorker.java
kmgowda: yes, currently I am just making sure that reader and transaction works fine but i am not publishing the performance numbers of reader and transactions.  once we close the pravega writer performance v/s kafka writer , I will revisit this code again.

PR #21, opened 2019-01-23
Refactor pravega writer and readers with the abstract Writers and readers

kmgowda:
The new abstract classes WriteWorker and ReadWorkers are created.
These classes does the performance bench-marking whereas the
pravega writer and pravega readers classes does the data writes and reads.


tkaitchuck:
This is based on an earlier PR #19 .

PR #22, opened 2019-01-24
Latency calculation update and percentile implementations

kmgowda:
The latency calculation is updated and latency percentile are implemented.

### DIFF
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -19,9 +19,9 @@ package com.emc.pravega.perf;
 
 
 
-import io.pravega.client.stream.AckFuture;
 
 import java.util.Arrays;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.function.Supplier;
 
@@ -145,15 +145,15 @@ class PerfStats {
         return values;
     }
 
-    public AckFuture runAndRecordTime(Supplier<AckFuture> fn, long startTime, int length, Executor executor) {
+    public CompletableFuture runAndRecordTime(Supplier<CompletableFuture> fn, long startTime, int length, Executor executor) {
         int iter = this.iteration++;
-        AckFuture retVal = fn.get();
+        CompletableFuture  retVal = fn.get();
         if(retVal == null) {
             record(iter, (int) (System.currentTimeMillis() - startTime) * 1000, length, System.nanoTime());
         } else {
-            retVal.addListener(() -> {
+            retVal = retVal.thenAccept((d) -> {
                 record(iter, (int) (System.currentTimeMillis() - startTime) * 1000, length, System.nanoTime());
-            }, executor);
+            });
         }
         return retVal;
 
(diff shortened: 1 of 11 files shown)


src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -110,24 +113,21 @@ public synchronized void printAll() {
      */
     public synchronized void printTotal(Instant endTime) {
         final long elapsed = Duration.between(start, endTime).toMillis();
-        double recsPerSec = 1000.0 * iteration / (double) elapsed;
+        double recsPerSec = 1000.0 * count / (double) elapsed;
```
tkaitchuck: This multiplying by `1000.0` could be moved onto elapsed, then it wouldn't need to be duplicated. The same pattern is in `printWindow()` above.

src/main/java/com/emc/pravega/perf/ReaderWorker.java
```diff
@@ -46,17 +46,17 @@
...
         String ret = null;
         try {
 
             for (int i = 0; (i < eventsPerWorker) &&
                     (Duration.between(StartTime, Instant.now()).getSeconds() < secondsToRun); i++) {
+                final Instant startTime = Instant.now();
                 ret = readData();
                 if (ret != null) {
-                    stats.recordTime(null, Instant.ofEpochMilli(Long.parseLong(ret.split(",")[0])), ret.length());
+                    stats.recordTime(null, startTime, ret.length());
```
tkaitchuck: This line is changing what is being measured. Before it was measuring writer->reader latency, now it is measuring the latency of `readData()`. Those are different, I want to check if this is intended.

src/main/java/com/emc/pravega/perf/PerfStats.java
kmgowda: Thanks.. yes, this done now.. elapsed will be in seconds now.

src/main/java/com/emc/pravega/perf/ReaderWorker.java
kmgowda: yes, for now.. it will be latency of reads ; once i published depending on the inputs , we may have write another type of producer /consumers for writer->reader latencies.

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -87,9 +87,9 @@ private synchronized void record(int bytes, Instant startTime, Instant endTime)
     }
 
     private void printWindow(Instant endTime) {
-        final long elapsed = Duration.between(windowStartTime, endTime).toMillis();
-        final double recsPerSec = 1000.0 * windowCount / (double) elapsed;
-        final double mbPerSec = 1000.0 * this.windowBytes / (double) elapsed / (1024.0 * 1024.0);
+        final long elapsed = Duration.between(windowStartTime, endTime).getSeconds();
```
tkaitchuck: Looking at the Javadoc, getSeconds() preforms a truncation to get the value in whole seconds. Given how narrow these windows are, I'm not sure we want this kind of inprecision. So let's make elapsed a double that is millis()/1000.0.

PR #23, opened 2019-02-07
events per second implementation

kmgowda:
"events per second" implemented if the time is supplied; otherwise events parameter is treated as events per producer/consumer.
shutdown hook is added to handle ctrl-c gracefully

### DIFF
--- a/src/main/java/com/emc/pravega/perf/PerfStats.java
+++ b/src/main/java/com/emc/pravega/perf/PerfStats.java
@@ -17,10 +17,17 @@
  */
 package com.emc.pravega.perf;
 
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.concurrent.CompletableFuture;
 import java.time.Instant;
 import java.time.Duration;
+import java.util.concurrent.atomic.AtomicLong;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVPrinter;
 
 
 /**
@@ -39,6 +46,8 @@ public class PerfStats {
     private long totalLatency;
     final private long windowInterval;
     private timeWindow window;
+    final private AtomicLong eventID;
+    final private CSVPrinter printer;
 
     /**
      * private class for Performance statistics within a given time window.
@@ -103,7 +112,7 @@ public class PerfStats {
         }
     }
 
-    public PerfStats(String action, int reportingInterval, int messageSize, long numRecords) {
+    public PerfStats(String action, int reportingInterval, int messageSize, long numRecords, String csvFile) throws IOException {
         this.action = action;
         this.start = Instant.now();
         this.count = 0;
@@ -115,9 +124,17 @@ public class PerfStats {
         this.windowInterval = reportingInterval;
         this.messageSize = messageSize;
         this.window = new timeWindow();
+        this.eventID = new AtomicLong();
+        if (csvFile != null) {
+            this.printer = new CSVPrinter(Files.newBufferedWriter(Paths.get(csvFile)), CSVFormat.DEFAULT
+                    .withHeader("#", "Event ID", "event size (bytes)", action + " Latency (milliseconds)"));
+        } else {
+            this.printer = null;
+        }
+
     }
 
-    private synchronized void record(int bytes, Instant startTime, Instant endTime) {
+    private synchronized void record(long event, int bytes, Instant startTime, Instant endTime) throws IOException {
 
         final long latency = Duration.between(startTime, endTime).toMillis();
         this.count++;
@@ -129,6 +146,10 @@ public class PerfStats {
         if (this.count % this.sampling == 0) {
             this.latencies[index] = latency;
             this.index++;
+        }
+
+        if (this.printer != null) {
+            printer.printRecord(count, event, bytes, latency);
         }
     }
 
@@ -146,19 +167,22 @@ public class PerfStats {
     /**
      * print the performance statistics of current time window.
      */
-    public synchronized void print() {
+    public synchronized void print() throws IOException {
 
         final Instant time = Instant.now();
         if (window.windowTimeMS(time) >= windowInterval) {
             window.print(time);
             this.window = new timeWindow();
+            if (printer != null) {
+                printer.flush();
+            }
         }
     }
 
     /**
      * print the final performance statistics.
      */
-    public synchronized void printTotal(Instant endTime) {
+    public synchronized void printTotal(Instant endTime) throws IOException {
 
         final double elapsed = Duration.between(start, endTime).toMillis() / 1000.0;
         final double recsPerSec = count / elapsed;
@@ -169,6 +193,9 @@ public class PerfStats {
                 "%d records %s, %.3f records/sec, %d bytes record size, %.2f MB/sec, %.1f ms avg latency, %.1f ms max latency, %d ms 50th, %d ms 95th, %d ms 99th, %d ms 99.9th.\n",
                 count, action, recsPerSec, messageSize, mbPerSec, totalLatency / ((double) count), (double) maxLatency,
                 percs[0], percs[1], percs[2], percs[3]);
+        if (printer != null) {
+            printer.close();
+        }
     }
 
     /**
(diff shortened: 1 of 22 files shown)


src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -55,33 +57,79 @@
...
-        return null;
+            while (Duration.between(StartTime, Instant.now()).getSeconds() < secondsToRun) {
+                final Instant loopStart = Instant.now();
+                long loopTime;
+
+                for (int i = 0; ((loopTime = Duration.between(loopStart, Instant.now()).toMillis()) < 1000) &&
+                        (i < events); i++) {
+
+                    // Construct event payload
+                    String val = System.currentTimeMillis() + ", " + workerID + ", " + (int) (Math.random() * 200);
```
RaulGracia: I don't think we need all this stuff for every write. We can have the `workerId` as routing key and a fixed payload of the desired size built outside of the loop (perhaps you could just append a sequenceId to the payload for verification purposes).

src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -55,33 +57,79 @@
...
+                    String payload = String.format("%-" + messageSize + "s", val);
+                    String key;
+                    if (isRandomKey) {
+                        key = Integer.toString(workerID + rand.nextInt());
+                    } else {
+                        key = Integer.toString(workerID);
+                    }
+
+                    final Instant startTime = Instant.now();
+                    retFuture = writeData(key, payload);
```
RaulGracia: As I understand, the purpose of this PR is to create a "rate limiter" of the events written in the system. While I could have missed something, I don't see anywhere in the code where are we preventing writers to write as fast as they can (I would expect a "sleep" or something similar to distribute writes along time based on some input rate).

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: Thanks Raul, But, I would like keep this because, every time i will write it the time as part of message so that , during reading i can get the full response time    (write-> read)

src/main/java/com/emc/pravega/perf/WriterWorker.java
kmgowda: after this loop, we sleep for 1000-looptime milliseconds

RaulGracia: I think we need a sleep after "each" write, otherwise you are generating bursts of requests that are not keeping a constant rate of events per second. And this will impact the latency of operations.

src/main/java/com/emc/pravega/perf/ReaderWorker.java
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- *
+ * <p>
```
tkaitchuck: Please don't alter these headers

src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -55,33 +57,79 @@
...
+                retFuture = stats.recordTime(retFuture, startTime, payload.length());
             }
 
-            final Instant startTime = Instant.now();
-            retFuture = writeData(key, payload);
-            // event ingestion
-            retFuture = stats.recordTime(retFuture, startTime, payload.length());
+            flush();
+
+            //Wait for the last packet to get acked
```
tkaitchuck: A call to flush on an event stream writer, will block until the last packet is acked.

src/main/java/com/emc/pravega/perf/WriterWorker.java
```diff
@@ -55,33 +57,82 @@
...
+                }
 
-        return null;
+                final Instant beginTime = Instant.now();
+                retFuture = writeData(key, payload);
+                // event ingestion
+                retFuture = stats.recordTime(retFuture, beginTime, payload.length());
+
+                if (sleep_ns > MIN_SLEEP_NS) {
+                    Thread.sleep(sleep_ms, (int) remain_ns);
```
tkaitchuck: Make sure you handle over-sleeping. The OS can't guarentee thread wake up time.

kmgowda: hi Tom, for oversleeping, I have done the time adjustment for next iteration ; but now its part of throughput controller

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -41,88 +39,107 @@
     private long bytes;
     private long maxLatency;
     private long totalLatency;
-    private long windowMaxLatency;
-    private long windowTotalLatency;
-    private long windowCount;
-    private long windowBytes;
-    final private long reportingInterval;
+    final private long windowInterval;
+    private timeWindow window;
```
tkaitchuck: @GuardedBy("lock") ?

kmgowda: Thanks Tom, But just personal choice, I have made all the public (Except one) as synchronized methods so that all the member variables stats are protected.

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -41,80 +39,132 @@
...
+        }
+
+        /**
+         * record the latency and bytes
+         *
+         * @param latency latency in ms.
+         * @param bytes   number of bytes.
+         */
+        public void record(long latency, long bytes) {
+            this.count++;
```
RaulGracia: In our discussions, we decided to have a raw file with the latency of operations. Maybe I'm missing something, but I'm not seeing the csv-like output of the operations. Maybe here, we need so write to a file all (or a evenly distributed fraction of) the operations:
```
Timestamp, Payload size, read/write, latency
1,100,0,20
2,100,0,23
...
``` 
Otherwise, we lose a lot of data to understand the behavior of latency (e.g., spike periodicity, configurable windows).

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -18,21 +18,19 @@
 package com.emc.pravega.perf;
 
 import java.util.Arrays;
-import java.util.ArrayList;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.Executor;
 import java.time.Instant;
 import java.time.Duration;
+import java.util.concurrent.locks.ReentrantLock;
```
RaulGracia: Is this being used somewhere?

src/main/java/com/emc/pravega/perf/PravegaPerfTest.java
```diff
@@ -173,24 +166,44 @@ public static void main(String[] args) {
...
+            shutdown();
         } catch (Exception e) {
             e.printStackTrace();
         }
 
         System.exit(0);
     }
 
+    private static synchronized void shutdown() throws InterruptedException {
+        final Instant endTime=Instant.now();
```
RaulGracia: Minor: Add an space before and after `=`. Maybe we need to import the checkstyle from Pravega repository to this one as well.

src/main/java/com/emc/pravega/perf/PravegaWriterWorker.java
```diff
@@ -20,28 +20,26 @@
...
  * class for Pravega writer/producer.
  */
 public class PravegaWriterWorker extends WriterWorker {
     final EventStreamWriter<String> producer;
 
-    PravegaWriterWorker(int sensorId, int eventsPerWorker, int secondsToRun,
+    PravegaWriterWorker(int sensorId, int events, int secondsToRun,
                         boolean isRandomKey, int messageSize, Instant start,
-                        PerfStats stats, String streamName, ClientFactory factory) {
+                        PerfStats stats, String streamName, ThroughputController tput,ClientFactory factory) {
```
RaulGracia: Minor: space needed after `,`.

src/main/java/com/emc/pravega/perf/PerfStats.java
kmgowda: removed now

src/main/java/com/emc/pravega/perf/PravegaWriterWorker.java
kmgowda: yes , I copied the checkstyle from pravega

src/main/java/com/emc/pravega/perf/PerfStats.java
kmgowda: yes, but I am thinking adding the metrics along with the csv reporter.. so, i think it should go as separate PR please.

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -115,9 +124,17 @@ public PerfStats(String action, int reportingInterval, int messageSize, long num
         this.windowInterval = reportingInterval;
         this.messageSize = messageSize;
         this.window = new timeWindow();
+        this.eventID = new AtomicLong();
+        if (csvFile != null) {
+            this.printer = new CSVPrinter(Files.newBufferedWriter(Paths.get(csvFile)), CSVFormat.DEFAULT
+                    .withHeader("#", "Event ID", "event size (bytes)", action + " Latency (milliseconds)"));
```
RaulGracia: Where is the timestamp? We need this to correlate with server side metrics.

src/main/java/com/emc/pravega/perf/PerfStats.java
```diff
@@ -149,7 +149,8 @@ private synchronized void record(long event, int bytes, Instant startTime, Insta
         }
 
         if (this.printer != null) {
-            printer.printRecord(count, event, bytes, latency);
+            final long nanotime = startTime.getEpochSecond() * 1000000000L + startTime.getNano();
```
RaulGracia: Please, do this conversion using a constant for `1000000000L` if there is no method that already does that for you.

PR #36, opened 2019-03-15
Performance benchmark tool improvements

### LINKED ISSUE #33: Improve the pravega benchmark tool burst more events
Currently the performance benchmark tool does not push the messages as kafka benchmark tool because of synchronized calls, Time instants are used instead of System.currentTimeMillis() and too many computations in the synchronized methods.
re design the tool to make sure that it pushes more and more events to pravega client.

kmgowda:
Change log description
Currently the performance benchmark tool does not push the messages as kafka benchmark tool because of synchronized calls, Time instants are used instead of System.currentTimeMillis() and too many computations in the synchronized methods. But, the pravega benchmark tool should support multi producer and multi consumers.  Hence , the pravega bench-marking tool is re designed to use the java concurrency. 

Purpose of the change
Fixes #33 .

What the code does
Java concurrency is used so that a dedicate processor implements the latency, throughput and CSV files related operations. hence need of synchronized methods are removed.
Time instants are replaced by System.currentTimeMillis().
end to end latency measurement are added
Documentation about how to build and use the pravega benchmark tool is
added.

How to verify it
build the tool and c producer and consumer to write and read the events to/from pravega cluster.
execute the end -end latency variant by supplying both producers and consumers.


src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -10,52 +10,61 @@
...
-    private long bytes;
-    private long maxLatency;
-    private long totalLatency;
+    final private String csvFile;
+    final private int messageSize;
     final private long windowInterval;
-    private timeWindow window;
-    final private AtomicLong eventID;
-    final private CSVPrinter printer;
+    private TimeWindow window;
```
tkaitchuck: This is being accessed in multiple threads and is non-final. See comments below highlighting the race condition.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -94,138 +103,245 @@ public void print(Instant time) {
...
-                "%d records %s, %.3f records/sec, %d bytes record size, %.2f MB/sec, %.1f ms avg latency, %.1f ms max latency, %d ms 50th, %d ms 95th, %d ms 99th, %d ms 99.9th.\n",
-                count, action, recsPerSec, messageSize, mbPerSec, totalLatency / ((double) count), (double) maxLatency,
-                percs[0], percs[1], percs[2], percs[3]);
-        if (printer != null) {
-            printer.close();
-        }
+    public synchronized void start(long startTime) throws IOException {
+        this.latencyRecorder = csvFile == null ? new LatencyWriter(action, messageSize, startTime) :
+                new CSVLatencyWriter(action, messageSize, startTime, csvFile);
+        this.window = new TimeWindow(startTime);
```
tkaitchuck: This is performed in an external caller's thread.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -94,138 +103,245 @@ public void print(Instant time) {
...
+        executor.execute(new ProcessQueue());
```
tkaitchuck: Oops this process queue is going to access window, but window was not passed through the queue or under any lock, so line 318 and 316 above might not have run yet from that thread's point of view.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -10,52 +10,61 @@
...
+    final private String csvFile;
+    final private int messageSize;
     final private long windowInterval;
-    private timeWindow window;
-    final private AtomicLong eventID;
-    final private CSVPrinter printer;
+    private TimeWindow window;
+    final private ConcurrentLinkedQueue<TimeStamp> queue;
+    final private ForkJoinPool executor;
+    private LatencyWriter latencyRecorder;
```
tkaitchuck: Same here.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -94,138 +103,245 @@ public void print(Instant time) {
...
+                }
+            }
         }
-        return values;
+    }
+
+
+    private void record(long startTime, long endTime, int bytes) {
+        final int latency = (int) (endTime - startTime);
+        window.record(bytes, latency);
```
tkaitchuck: This occurs the the executor thread with no lock held. So an NPE could be thrown right here.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -94,138 +103,245 @@ public void print(Instant time) {
...
+        latencyRecorder.record(startTime, bytes, latency);
```
tkaitchuck: same here

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -94,138 +103,245 @@ public void print(Instant time) {
...
     /**
      * print the performance statistics of current time window.
      */
-    public synchronized void print() throws IOException {
-
-        final Instant time = Instant.now();
+    private void print() throws IOException {
+        final long time = System.currentTimeMillis();
         if (window.windowTimeMS(time) >= windowInterval) {
             window.print(time);
```
tkaitchuck: Same here

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -94,138 +103,245 @@ public void print(Instant time) {
...
-        long[] percs = percentiles(this.latencies, index, 0.5, 0.95, 0.99, 0.999);
-        System.out.printf(
-                "%d records %s, %.3f records/sec, %d bytes record size, %.2f MB/sec, %.1f ms avg latency, %.1f ms max latency, %d ms 50th, %d ms 95th, %d ms 99th, %d ms 99.9th.\n",
-                count, action, recsPerSec, messageSize, mbPerSec, totalLatency / ((double) count), (double) maxLatency,
-                percs[0], percs[1], percs[2], percs[3]);
-        if (printer != null) {
-            printer.close();
-        }
+    public synchronized void start(long startTime) throws IOException {
+        this.latencyRecorder = csvFile == null ? new LatencyWriter(action, messageSize, startTime) :
```
tkaitchuck: This is performed in an external caller's thread

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -44,36 +40,39 @@
  * Data format is in comma separated format as following: {TimeStamp, Sensor Id, Location, TempValue }.
  */
 public class PravegaPerfTest {
-
+    private static final int maxTime = 60 * 60 * 24;
     private static String controllerUri = "tcp://localhost:9090";
```
tkaitchuck: All these need to be final.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -44,36 +40,39 @@
...
+    private static int events = 0;
+    private static int eventsPerSec = 0;
+    private static int eventsPerWorker = 0;
     private static int transactionPerCommit = 0;
     private static int runtimeSec = 0;
     private static final int reportingInterval = 5000;
     private static ScheduledExecutorService bgexecutor;
     private static ForkJoinPool fjexecutor;
-    private static PerfStats produceStats, consumeStats;
+    private static PerfStats produceStats = null;
```
tkaitchuck: This is by definition a race condition.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -44,36 +40,39 @@
...
+    private static PerfStats consumeStats = null;
     private static double throughput = 0;
-    private static ThroughputController tput;
     private static String writeFile = null;
     private static String readFile = null;
+    private static ReaderGroup readerGroup = null;
+    private static final int timeout = 10;
+    private static ClientFactory factory = null;
+    private static ControllerImpl controller = null;
+    private static List<Callable<Void>> readers = null;
```
tkaitchuck: A static mutable list that isn't final with callables (to be executed in threads) is 3 race conditions in 1 line! 
Make this, and all the other variables thread safe. (Preferably by making them instance variables.)

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -57,67 +74,116 @@ public Void call() throws InterruptedException, ExecutionException, IOException
...
-                    key = Integer.toString(workerID);
-                }
+    private class EventsController {
+        private static final long NS_PER_MS = 1000000L;
+        private static final long NS_PER_SEC = 1000 * NS_PER_MS;
+        private static final long MIN_SLEEP_NS = 2 * NS_PER_MS;
+        private final long startTime;
+        private final long sleepTimeNs;
+        private final int eventsPerSec;
+        private long toSleepNs  = 0;
```
tkaitchuck: What's going on here? This is clearly not final, it's being modified from a public method, There is no annotation declaring where the lock is. How is anyone supposed to verify if this is ok or not?

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: now, this  timewindow and latency recorder are as part of QueueProcessor object.

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: yes, I have changed the implementation of start and shutdown methods

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: ok, now the constructor of the ProcessQueue creates the timewindow and latency recorder

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: yes, so, I have changed so that  ProcessQueue creates the timewindow and latency recorder

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: this record is always called by dedicated thread;  multiple threads wont access this method.

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: controllerUri  can change depending on the user input

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: I have refactored it to use it as local final variables.

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: now, I have tagged EventsController as  NotThreadSafe;
however, each producer will have its own EventsController  object, so no multiple threads/writers accessing the single EventsController  .

src/main/java/io/pravega/perf/PravegaPerfTest.java
tkaitchuck: Then it cannot be static. Otherwise you have a race condition.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -10,39 +10,57 @@
...
+                 PerfStats stats, String streamName, int eventsPerSec, boolean wNr) {
 
         super(sensorId, events, secondsToRun,
-                isRandomKey, messageSize, start,
-                stats, streamName, 0);
-        this.tput = tput;
-        perf = secondsToRun > 0 ? new ThroughputWriter() : new EventsWriter();
+                messageSize, start, stats,
+                streamName, 0);
+        this.eCnt = new EventsController(start, eventsPerSec);
```
tkaitchuck: eCnt is not a thread safe class. It is being constructed here in one thread, and benchmark() is being invoked by another. No class should depend on it's callers to be correct, as it will lead to bugs like this one. Instead eCnt should be made into a local variable instantiated at the top of benchmark(). This will guarantee it is only used by one thread.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -10,39 +10,57 @@
...
+        perf = secondsToRun > 0 ? (wNr ? new EventsWriterTimeRW() : new EventsWriterTime()) :
+                (wNr ? new EventsWriterRW() : new EventsWriter());
+
+        Random random = new Random(0);
+        byte[] bytes = new byte[messageSize];
+        for (int i = 0; i < messageSize; ++i) {
+            bytes[i] = (byte) (random.nextInt(26) + 65);
+        }
+        payload = new String(bytes);
```
tkaitchuck: Specify the character encoding so this does not generate a warning.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -57,67 +75,117 @@ public Void call() throws InterruptedException, ExecutionException, IOException
...
-                String val = System.currentTimeMillis() + ", " + workerID + ", " + (int) (Math.random() * 200);
-                String payload = String.format("%-" + messageSize + "s", val);
-                String key;
-                if (isRandomKey) {
-                    key = Integer.toString(workerID + rand.nextInt());
-                } else {
-                    key = Integer.toString(workerID);
-                }
+    @NotThreadSafe
+    private class EventsController {
```
tkaitchuck: static

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,389 @@
...
-    final private AtomicLong eventID;
-    final private CSVPrinter printer;
+    final private ConcurrentLinkedQueue<TimeStamp> queue;
+    final private ForkJoinPool executor;
+    private QueueProcessor recorder;
+
+    /**
+     * private class for start and end time.
+     */
+    @NotThreadSafe
```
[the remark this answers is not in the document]
kmgowda: yes, it changed now

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,389 @@
...
-    private long totalLatency;
+    final private String csvFile;
+    final private int messageSize;
     final private long windowInterval;
-    private timeWindow window;
-    final private AtomicLong eventID;
-    final private CSVPrinter printer;
+    final private ConcurrentLinkedQueue<TimeStamp> queue;
+    final private ForkJoinPool executor;
+    private QueueProcessor recorder;
```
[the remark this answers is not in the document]
kmgowda: please guide me here.. QueueProcessor  procesor class has the its own thread; except shutdown -> printTotal (which is final action) , its members are accessed by it own thread only.
The start method creates the recorder object and shutdown method destroys this object.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,389 @@
...
-         * @param latency latency in ms.
-         * @param bytes   number of bytes.
-         */
-        public void record(long latency, long bytes) {
-            this.count++;
-            this.totalLatency += latency;
-            this.bytes += bytes;
-            this.maxLatency = Math.max(this.maxLatency, latency);
+        private void shutdown(long endTime) throws ExecutionException, InterruptedException {
+            queue.add(new TimeStamp(endTime));
```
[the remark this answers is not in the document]
kmgowda: ok, now.. its changed so that.. queue.add is only by record time.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,389 @@
...
-            this.startTime = Instant.now();
-            this.lastTime = this.startTime;
-            this.count = 0;
-            this.bytes = 0;
-            this.maxLatency = 0;
-            this.totalLatency = 0;
+    @NotThreadSafe
+    final private class QueueProcessor implements Runnable {
+        final private LatencyWriter latencyRecorder;
+        final private TimeWindow window;
```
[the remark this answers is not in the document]
kmgowda: yes, its done now.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,389 @@
...
+    }
 
 
+    public PerfStats(String action, int reportingInterval, int messageSize, String csvFile) {
+        this.action = action;
+        this.messageSize = messageSize;
+        this.windowInterval = reportingInterval;
+        this.csvFile = csvFile;
+        this.queue = new ConcurrentLinkedQueue<TimeStamp>();
+        this.executor = new ForkJoinPool(1);
```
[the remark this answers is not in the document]
kmgowda: yes, its done

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,389 @@
...
         /**
-         * get the time duration of this window
+         * private class for Performance statistics within a given time window.
          */
-        public long windowTimeMS() {
-            return windowTimeMS(Instant.now());
-        }
-    }
+        @NotThreadSafe
+        private class TimeWindow {
```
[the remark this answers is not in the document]
kmgowda: yes, its done

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,389 @@
...
-    }
 
-    private static long[] percentiles(long[] latencies, int count, double... percentiles) {
-        int size = Math.min(count, latencies.length);
-        Arrays.sort(latencies, 0, size);
-        long[] values = new long[percentiles.length];
-        for (int i = 0; i < percentiles.length; i++) {
-            int index = (int) (percentiles[i] * size);
-            values[i] = latencies[index];
+        private class LatencyWriter {
```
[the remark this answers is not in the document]
kmgowda: yes, its done

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: yes, its done.. now its local variable instantiated  just before the bench marking starts.

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: its changed to US_ASCII.

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: yes, it done

src/main/java/io/pravega/perf/PerfStats.java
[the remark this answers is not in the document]
kmgowda: i dont want use any locks here because QueueProcessor  has it own thread to update its member variables.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,389 @@
...
-         * @param bytes   number of bytes.
-         */
-        public void record(long latency, long bytes) {
-            this.count++;
-            this.totalLatency += latency;
-            this.bytes += bytes;
-            this.maxLatency = Math.max(this.maxLatency, latency);
+        private void shutdown(long endTime) throws ExecutionException, InterruptedException {
+            queue.add(new TimeStamp(endTime));
+            this.end.get();
```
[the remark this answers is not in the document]
kmgowda: yes, its removed now

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: no threads access this .. its default and one time assignment.

src/main/java/io/pravega/perf/PerfStats.java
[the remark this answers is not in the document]
tkaitchuck: There are two different issues here. The QueueProcessor itself. (See my comment below) and this member is non-final. This is ok if it guarded by a lock. It looks to be guarded by 'this' so it should be annotated accordingly.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,227 +5,363 @@
...
+        final private long windowInterval;
+        final private ConcurrentLinkedQueue<TimeStamp> queue;
+
+
+        private QueueProcessor(String action, int reportingInterval, int messageSize,
+                               long startTime, String csvFile, ConcurrentLinkedQueue<TimeStamp> queue) throws IOException {
+            this.action = action;
+            this.windowInterval = reportingInterval;
+            this.queue = queue;
+            this.latencyRecorder = csvFile == null ? new LatencyWriter(action, messageSize, startTime) :
```
tkaitchuck: If this instantiation were delayed until run(), and instead 'action, messageSize, and startTime' were final member variables, then this class would be thread safe.

src/main/java/io/pravega/perf/PravegaPerfTest.java
tkaitchuck: You are correct. This means all of these members can be removed altogether. They are only ever used by the same method that assigned them. So they can be local variables.

src/main/java/io/pravega/perf/PerfStats.java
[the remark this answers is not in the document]
kmgowda: yes , I have added the guarded by "this" for "recorder" member variable

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: ok, its done

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: I have refactored the this complete file. please check.

src/main/java/io/pravega/perf/PerfStats.java
tkaitchuck: I do not see this change.

kmgowda: now , I have made the Latency writer as a local variable to queue processor thread.
to stop this thread, I will be sending end msg to queue.   I remember that last time you commented  that a sending end msg looks odd; but , I need full latency calculations at the end of the bench-marking and i dont want to use locking here . so, this approach should be ok with respect to achieving thread safety by not using locks.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -289,23 +242,180 @@ private static void parseCmdLine(String[] args) throws ParseException {
...
-        static final String SCOPE = "Scope";
-        static final String STREAM_NAME = "aaj";
+    static private class PravegaTest extends Test {
+        final PravegaStreamHandler streamHandle;
+        final ClientFactory factory;
+
+        PravegaTest(long startTime, CommandLine commandline) throws
+                IllegalArgumentException, URISyntaxException, InterruptedException, Exception {
+            super(startTime, commandline);
+            final ScheduledExecutorService bgExecutor = Executors.newScheduledThreadPool(10);
```
tkaitchuck: Is 10 chosen for a reason?

kmgowda: Thanks Tom.. No , it its existing since quite some time.. please let me know if you have any preference value for this.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -94,138 +182,197 @@ public void print(Instant time) {
...
-    public synchronized void print() throws IOException {
-
-        final Instant time = Instant.now();
-        if (window.windowTimeMS(time) >= windowInterval) {
-            window.print(time);
-            this.window = new timeWindow();
-            if (printer != null) {
-                printer.flush();
+    @NotThreadSafe
+    static private class CSVLatencyWriter extends LatencyWriter {
```
RaulGracia: How do you guarantee that this class is used in a thread-safe manner?

src/main/java/io/pravega/perf/PravegaReaderWorker.java
```diff
@@ -5,42 +5,37 @@
...
 public class PravegaReaderWorker extends ReaderWorker {
     private final EventStreamReader<String> reader;
-    private final String readerId;
 
     PravegaReaderWorker(int readerId, int events, int secondsToRun,
-                        Instant start, PerfStats stats, String readergrp,
-                        int timeout, ClientFactory factory) {
-        super(readerId, events, secondsToRun, start, stats, readergrp, timeout);
+                        long start, PerfStats stats, String readergrp,
+                        int timeout, boolean wNr, ClientFactory factory) {
```
RaulGracia: Please, choose a self-explanatory name for `wNr`.

README.md
```diff
@@ -1,4 +1,170 @@
...
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+-->
+
+# Pravega Benchmark Tool
+
+The pravega benchmark tool used for the performance benchmarking of pravega streaming storage cluster.
```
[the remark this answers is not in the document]
kmgowda: yes done.

README.md
```diff
@@ -1,4 +1,170 @@
...
+This tool performs the throughput and latency analysis for the multi producers and consumers of pravega.
```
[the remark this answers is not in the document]
kmgowda: help and everything has producers and consumers, so I updated it as producers/writers 
consumers/readers

README.md
```diff
@@ -1,4 +1,170 @@
...
+it also validates the end to end latency. The write and/or read latencies can be stored in the CSV file for later analysis.
+At the end of the performance benchmarking, this tool outputs the 50th, 75th, 95th , 99th and 99.9th latency percentiles.
+
+
+### Prerequisites
+
+- Pravega client : https://github.com/pravega/pravega
```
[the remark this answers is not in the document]
kmgowda: removed

README.md
```diff
@@ -1,4 +1,170 @@
...
+For example:
+<pravega benchmark directory>/run/pravega-benchmark/bin/pravega-benchmark  -controller tcp://127.0.0.1:9090  -stream streamname1  -segments 1  -producers 1  -size 100   -throughput -1   -time 60
+
+The -throughput -1  indicates the burst mode.
+This test will executed for 60 seconds because option -time 60 is used.
+This test tries to write and read events of size 100 bytes to/from the stream 'streamname1'.
+The option '-controller tcp://127.0.0.1:9090' specifies the pravega controller IP address and port number.
+Note that -producers 1 indicates 1 producer/writers.
+
+in case you want to write/read the certain number of events use the -events option without -time option as follows
```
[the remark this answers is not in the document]
kmgowda: done

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -79,9 +167,9 @@ public void record(long latency, long bytes) {
         /**
          * print the window statistics
          */
-        public void print(Instant time) {
+        private void print(long time) {
             this.lastTime = time;
-            final double elapsed = Duration.between(this.startTime, this.lastTime).toMillis() / 1000.0;
+            final double elapsed = (this.lastTime - this.startTime) / 1000.0;
```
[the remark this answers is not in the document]
kmgowda: yes , its added

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: CSVLatencyWriter  is used as local variable inside the while call method the queue processor

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,57 +5,145 @@
...
+        public Void call() throws IOException {
+            final TimeWindow window = new TimeWindow(action, startTime);
+            final LatencyWriter latencyRecorder = csvFile == null ? new LatencyWriter(action, messageSize, startTime) :
+                    new CSVLatencyWriter(action, messageSize, startTime, csvFile);
+            long endTime = 0;
+            boolean yes = true;
+            long time;
+            TimeStamp t;
+
+            while (yes) {
```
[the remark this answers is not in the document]
kmgowda: changed as "doWork"

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,57 +5,145 @@
...
+        this.executor = new ForkJoinPool(1);
+        this.ret = null;
+    }
+
+    /**
+     * private class for start and end time.
+     */
+    final private class QueueProcessor implements Callable {
+        final private long startTime;
+
```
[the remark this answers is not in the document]
kmgowda: removed

README.md
```diff
@@ -1,4 +1,170 @@
...
+
+<pravega benchmark directory>/run/pravega-benchmark/bin/pravega-benchmark  -controller tcp://127.0.0.1:9090  -stream streamname1  -segments 1  -producers 1  -size 100   -throughput -1   -events 1000000
+
+-events <number> indicates that total <number> of events to write/read
+```
+
+### 2 - Throughput Mode
+In this mode, the pravega benchmark tool pushes the messages to the pravega client with specified approximate maximum throughput.
+This mode is used to find the least latency  that can be obtained from the pravega cluster for given throughput.
```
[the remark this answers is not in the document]
kmgowda: removed

README.md
```diff
@@ -1,4 +1,170 @@
...
+Note that -producers 1 indicates 1 producer/writers.
+
+in case you want to write/read the certain number of events use the -events option without -time option as follows
+
+<pravega benchmark directory>/run/pravega-benchmark/bin/pravega-benchmark  -controller tcp://127.0.0.1:9090  -stream streamname1  -segments 1  -producers 1  -size 100   -throughput -1   -events 1000000
+
+-events <number> indicates that total <number> of events to write/read
+```
+
+### 2 - Throughput Mode
```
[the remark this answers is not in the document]
kmgowda: yes, I have updated the same

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,57 +5,145 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
```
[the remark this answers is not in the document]
kmgowda: yes, changed with tabs

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,57 +5,145 @@
...
+        private TimeStamp(long endTime) {
+            this(-1, endTime, -1);
+        }
+
+        private boolean isEnd() {
+            return this.bytes == -1 && this.startTime == -1;
+        }
+
+    }
+
```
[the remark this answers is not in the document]
kmgowda: removed

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,57 +5,145 @@
...
+    final private int messageSize;
+    final private int windowInterval;
+    final private ConcurrentLinkedQueue<TimeStamp> queue;
+    final private ForkJoinPool executor;
+
+    @GuardedBy("this")
+    private Future<Void> ret;
+
+    /**
+     * private class for start and end time.
```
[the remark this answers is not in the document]
kmgowda: yes.. taken care all the comments.

README.md
```diff
@@ -1,4 +1,170 @@
...
+The -throughput <positive number>  indicates the Throughput mode.
+
+This test will be executed with approximate max throughput of 10MB/sec.
+This test will executed for 300 seconds (5 minutes) because option -time 60 is used.
+This test tries to write and read events of size 100 bytes to/from the stream 'streamname5' of 5 segments.
+If the stream 'streamname5' is not existing , then it will be created with the 5 segments.
+if the steam is already existing then it will be scaled up/down to 5 segments.
+Note that -producers 5 indicates 5 producers/writers .
+
+in case you want to write/read the certain number of events use the -events option without -time option as follows
```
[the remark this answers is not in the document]
kmgowda: done

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -5,57 +5,139 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
RaulGracia: Extra space.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -5,162 +5,112 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
RaulGracia: Extra space.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -5,47 +5,64 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
RaulGracia: Check this extra space in this specific modified line across files, there are several.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -5,47 +5,64 @@
...
     final private Performance perf;
-    final private ThroughputController tput;
+    final private String payload;
+    final private int eventsPerSec;
 
     WriterWorker(int sensorId, int events, int secondsToRun,
-                 boolean isRandomKey, int messageSize, Instant start,
-                 PerfStats stats, String streamName, ThroughputController tput) {
+                 boolean isRandomKey, int messageSize, long start,
+                 PerfStats stats, String streamName, int eventsPerSec, boolean writeNread) {
```
RaulGracia: `writeNread` -> means `writeAndRead`, right? Maybe `writeAndRead` would be easier to understand. Otherwise, it can be confused to write `N` events, and then read.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -5,47 +5,64 @@
...
+        byte[] bytes = new byte[messageSize];
+        for (int i = 0; i < messageSize; ++i) {
+            bytes[i] = (byte) (random.nextInt(26) + 65);
+        }
+        payload = new String(bytes, StandardCharsets.US_ASCII);
     }
 
-    /**
-     * writes the data.
+    /**.
```
RaulGracia: Please, remove this `.`.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -57,67 +74,120 @@ public Void call() throws InterruptedException, ExecutionException, IOException
...
+        }
+    }
 
+    private class EventsWriterRW implements Performance {
+
+        public void benchmark() throws InterruptedException, IOException {
+            final EventsController eCnt = new EventsController(System.currentTimeMillis(), eventsPerSec);
             for (int i = 0; i < events; i++) {
+                final String val = System.currentTimeMillis() + ", " + workerID + ", ";
+                final String data = (val + payload).substring(0, messageSize);
```
RaulGracia: You can get the `payload` of the appropriate size outside the loop, as the format of the String `val` is constant. This will make each iteration more efficient, as you could avoid executing `substring()`, just appending the `val` to the to the `payload`. Moreover, you could also add an assertion to verify that the payload size is the expected one.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -57,67 +74,120 @@ public Void call() throws InterruptedException, ExecutionException, IOException
...
+
+        public void benchmark() throws InterruptedException, IOException {
+            final long msToRun = secondsToRun * MS_PER_SEC;
+            long time = System.currentTimeMillis();
+            final EventsController eCnt = new EventsController(time, eventsPerSec);
+
+            for (int i = 0; (time - startTime) < msToRun; i++) {
+                time = System.currentTimeMillis();
+                final String val = time + ", " + workerID + ", ";
+                final String data = (val + payload).substring(0, messageSize);
```
RaulGracia: Similar comment to the previous one regarding the payload.

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: done

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: yes, changed in all the files

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: yes. its removed

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: yes , its renamed as  writeAndRead

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: yes, i have done the payload contruction outside the loop for methods EventsWriter , EventsWriterTime  but you are seeing the methods EventsWriterRW and EventsWriterTimeRW which are required for end to end latency.  for end to end latency, the start time will be existing in every event sent and so that reader reads its and marks it as a start time.

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: yes, its same as above.

README.md
```diff
@@ -1,4 +1,170 @@
...
+
+```
+<dir>/pravega-benchmark$ ./run/pravega-benchmark/bin/pravega-benchmark  -help
+usage: pravega-benchmark
+ -consumers <arg>               number of consumers
+ -controller <arg>              controller URI
+ -events <arg>                  number of events/records if 'time' not
+                                specified;
+                                otherwise, maximum events per second by
+                                producer(s)and/or number of events per
```
RaulGracia: Missing space after `producer(s)`.

src/main/java/io/pravega/perf/WriterWorker.java
RaulGracia: My point is not related to the fact that you need to calculate the timestamp for each message sent to get the end-to-end latency. Instead, I wonder about if we could make more efficient this line:
`final String data = (val + payload).substring(0, messageSize);`
This is doing a string concatenation and a call to `substring()` method per iteration, which is expensive. I suggest to explore the use `StringBuffer` for concatenation and avoid to operate as much as possible on `payload` within that loop as it may be up to 1MB in size.

README.md
kmgowda: yes correcting in readme file and code also.

kmgowda:
now, I have used the StringBuilder to construct the payload for end to end latency. only the hearder is changed for every write.

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: now, I have used the StringBuilder to construct the payload for end to end latency. only the hearder is changed for every write.

kmgowda:
Thanks Tom and Raul..

PR #40, opened 2019-04-17
Kafka benchmarking Implementation

### LINKED ISSUE #39: Extend the Pravega benchmark tool for Kafka performance benchmarking
Problem description
Extend the Pravega benchmark tool do the kafka bench-marking too.

Problem location
PravegaPerfTest

Suggestions for an improvement
make sure that --kafka option is added.  the same data set of pravega can be used to kafka bench marking too

kmgowda:
Change log description
The existing pravega bench marking tool is extended to do the Kafka bench marking.

Purpose of the change
Fixes #39 .

What the code does
The new classes KafkaWriterWorker and KafkaReaderWorker are added to kafka write, read and end to end latency bench marking.
Documentation about how to build and use the pravega benchmark tool for kafka bench marking is added.

How to verify it
build the tool;  use option "kafka -true" for Kafka bench marking with producers and consumers to write and read the events to/from Kafka cluster. Execute the end -end latency variant by supplying both producers and consumers.

Signed-off-by: Keshava Munegowda keshava.munegowda@dell.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ You may obtain a copy of the License at
 
 # Pravega Benchmark Tool
 
-The Pravega benchmark tool used for the performance benchmarking of pravega streaming storage cluster.
+The Pravega benchmark tool used for the performance benchmarking of pravega and Kafka streaming storage clusters.
 This tool performs the throughput and latency analysis for the multi producers/writers and consumers/readers of pravega.
 it also validates the end to end latency. The write and/or read latencies can be stored in a CSV file for later analysis.
 At the end of the performance benchmarking, this tool outputs the 50th, 75th, 95th , 99th and 99.9th latency percentiles.
@@ -55,6 +55,7 @@ usage: pravega-benchmark
                                 producer(s) and/or number of events per
                                 consumer
  -help                          Help message
+ -kafka <arg>                   Kafka Benchmarking
  -producers <arg>               number of producers
  -readcsv <arg>                 csv file to record read latencies
  -recreate <arg>                If the stream is already existing, delete
@@ -166,5 +167,9 @@ The -throughput -1 specifies the writes tries to write the events at the maximum
 ```
 
 ### Recording the latencies to CSV files
-user can use the options "-writecsv  <file name>" to record the latencies of writers and "-readcsv <file name>" for readers.
+User can use the options "-writecsv  <file name>" to record the latencies of writers and "-readcsv <file name>" for readers.
 in case of End to End latency mode, if the user can supply only -readcsv to get the end to end latency in to the csv file.
+
+### Kafka Benchmarking
+User can set the option "-kafka true"  for Kafka Benchmarking. User should create the topics manually before running this for kafka benchmarking.
+Unlike pravega benchmarking, this tool does not create the topic automatically. This tools treats stream name as a topic name.
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ usage: pravega-benchmark
  -producers <arg>               number of producers
  -readcsv <arg>                 csv file to record read latencies
  -recreate <arg>                If the stream is already existing, delete
-                                it and recreate it
+                                it and recreate it (not applicable for Kafka)
  -segments <arg>                Number of segments
  -size <arg>                    Size of each message (event or record)
  -stream <arg>                  Stream name
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ You may obtain a copy of the License at
 
 # Pravega Benchmark Tool
 
-The Pravega benchmark tool used for the performance benchmarking of pravega and Kafka streaming storage clusters.
+The Pravega benchmark tool used for the performance benchmarking of Pravega and Kafka streaming storage clusters.
 This tool performs the throughput and latency analysis for the multi producers/writers and consumers/readers of pravega.
 it also validates the end to end latency. The write and/or read latencies can be stored in a CSV file for later analysis.
 At the end of the performance benchmarking, this tool outputs the 50th, 75th, 95th , 99th and 99.9th latency percentiles.
@@ -88,8 +88,8 @@ The Pravega benchmark tool can be executed in the following modes:
 ```
 
 ### 1 - Burst Mode
-In this mode, the Pravega benchmark tool pushes/pulls the messages to/from the pravega client as much as possible.
-This mode is used to find the maximum and throughput that can be obtained from the pravega cluster.
+In this mode, the Pravega benchmark tool pushes/pulls the messages to/from the Pravega client as much as possible.
+This mode is used to find the maximum and throughput that can be obtained from the Pravega cluster.
 This mode can be used for both producers and consumers.
 
 ```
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -166,5 +167,9 @@ The -throughput -1 specifies the writes tries to write the events at the maximum
 ```
 
 ### Recording the latencies to CSV files
-user can use the options "-writecsv  <file name>" to record the latencies of writers and "-readcsv <file name>" for readers.
+User can use the options "-writecsv  <file name>" to record the latencies of writers and "-readcsv <file name>" for readers.
 in case of End to End latency mode, if the user can supply only -readcsv to get the end to end latency in to the csv file.
```
RaulGracia: After a dot, words should start in capitals.

README.md
```diff
@@ -166,5 +167,9 @@ The -throughput -1 specifies the writes tries to write the events at the maximum
...
+
+### Kafka Benchmarking
+User can set the option "-kafka true"  for Kafka Benchmarking. User should create the topics manually before running this for kafka benchmarking.
+Unlike pravega benchmarking, this tool does not create the topic automatically. This tools treats stream name as a topic name.
```
RaulGracia: If "Kafka" and "Pravega" are written with caps before, we need to keep the style consistent everywhere else.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -414,4 +430,88 @@ private void shutdown(long endTime) {
...
+
+        private Properties createProducerConfig() {
+            if (producerCount < 1) {
+                return null;
+            }
+            final Properties props = new Properties();
+            props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, controllerUri);
+            props.put(ProducerConfig.ACKS_CONFIG, "all");
+            props.put(ProducerConfig.RETRIES_CONFIG, 0);
+            props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName());
```
RaulGracia: Are the serializers being used for Pravega and Kafka equivalent? It would be good to use the same type of serializer, given that it would yields the same serialization overhead for both systems, which is fair.

src/main/java/io/pravega/perf/KafkaReaderWorker.java
```diff
@@ -0,0 +1,46 @@
...
+        this.consumer.subscribe(Arrays.asList(partition));
+    }
+
+    @Override
+    public String readData() {
+        final ConsumerRecords<String, String> records = consumer.poll(timeout);
+        if (records.isEmpty()) {
+            return null;
+        }
+        return records.iterator().next().value();
```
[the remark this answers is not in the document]
RaulGracia: Is this related to `props.put(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, 1);`? Is this an efficient configuration?

src/main/java/io/pravega/perf/ReaderWorker.java
```diff
@@ -69,12 +69,13 @@ public void benchmark() throws IOException {
         public void benchmark() throws IOException {
             String ret = null;
             try {
-                for (int i = 0; i < events; i++) {
+                for (int i = 0; i < events; ) {
```
RaulGracia: For clarity, maybe this needs to be a `while`.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -101,9 +101,16 @@ public void benchmark() throws InterruptedException, IOException {
...
                 final String data = buffer.replace(0, TIME_HEADER_SIZE, header).toString();
                 writeData(data);
+                /*
+                flush is required here for following reasons:
+                1. The writeData is called for End to End latency mode; hence make sure that data is sent.
+                2. In case of kafka benchmarking, the buffering makes too many writes;
+                   flushing moderates the kafka producer.
+                3. If the flush called after several iterations, then flush may take too much of time.
+                */
+                flush();
```
RaulGracia: So, this means that we are flushing for both Pravega and Kafka in the same way, right? This is only needed if you want to measure end-to-end latency, so if you execute a write-only workload you are not going to flush for every event? 
Another question: does it make sense that the invocation of `flush()` after every write could be configurable? For instance, in `fio` benchmark you can use `buffered` or  `direct` IO. Similarly, you may want to execute a benchmark with the buffering the system uses by default and flushing after every write as a config parameter in the benchmark.

README.md
```diff
@@ -10,7 +10,7 @@ You may obtain a copy of the License at
 
 # Pravega Benchmark Tool
 
-The Pravega benchmark tool used for the performance benchmarking of pravega streaming storage cluster.
+The Pravega benchmark tool used for the performance benchmarking of pravega and Kafka streaming storage clusters.
```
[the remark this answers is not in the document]
kmgowda: do i need to remove kafka here?

README.md
```diff
@@ -55,10 +55,11 @@ usage: pravega-benchmark
                                 producer(s) and/or number of events per
                                 consumer
  -help                          Help message
+ -kafka <arg>                   Kafka Benchmarking
```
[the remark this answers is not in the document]
kmgowda: yes.. for kafka we need to set this as "-kafka true"

src/main/java/io/pravega/perf/KafkaReaderWorker.java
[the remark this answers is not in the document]
kmgowda: yes, this same as pravega configuration ;  for experiment sake i did with multiple records.. after receiving the multiple records we need to process it any array.. so, reading single record is good here and it does not affect any performance.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -414,4 +430,88 @@ private void shutdown(long endTime) {
...
+        }
+
+        private Properties createProducerConfig() {
+            if (producerCount < 1) {
+                return null;
+            }
+            final Properties props = new Properties();
+            props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, controllerUri);
+            props.put(ProducerConfig.ACKS_CONFIG, "all");
+            props.put(ProducerConfig.RETRIES_CONFIG, 0);
```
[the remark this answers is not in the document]
kmgowda: ok, I am not setting the ProducerConfig.RETRIES_CONFIG now , it will have the default value of Integer MAX.
but , enable.idempotence is required for transactions so I will raise the separate PR for kafka transactions support and comparison with Pravega.

README.md
kmgowda: done

README.md
```diff
@@ -10,7 +10,7 @@ You may obtain a copy of the License at
 
 # Pravega Benchmark Tool
 
-The Pravega benchmark tool used for the performance benchmarking of pravega streaming storage cluster.
```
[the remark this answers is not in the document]
kmgowda: done

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: yes.. its same as UTF8String sterilizer used in pravega and for Kafka too: https://kafka.apache.org/0100/javadoc/org/apache/kafka/common/serialization/StringSerializer.html

src/main/java/io/pravega/perf/ReaderWorker.java
kmgowda: done

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: yes.. its in same way now.. 
for write only.. it does not flush every event.

for flushing for every write is possible , but, confirm if its required so that i will create another PR for implementing the same.

RaulGracia: In write-only mode, when the client is flushing? Apparently, it looks that after it writes all the events, but maybe I'm wrong. If this is the case, this may need to be configurable (e.g., after how many writes the writer flushes). This may be interesting to analyze the impact of flushing on latency and throughput.

kmgowda: Hi Raul, In write only mode also, now I have implemented "-flush" option in which you can supply the number of events per producer to flush. I have tested with both kafka and pravega.

Issue #41, opened 2019-05-13
add -flush option for performance bench-marking


kmgowda:
### Problem description
during writer performance bench marking , currently we flush the events once all the events are written or in end to end latency , every message is flushed. But, we need an option to specify by the user so that after "N" messages , flush should be invoked.

### Problem location
PravegaPerfTest

### Suggestions for an improvement
Add  "-flush <N>"  parameter to the pravega benchmark .
where N is number of messages to flush per producer. 
This option can be used to flush the message by producers while measuring write performance.

PR #42, opened 2019-05-15
Closing Readergroups, writers and adding "-flush" parameter to Benchmark tool

### LINKED ISSUE #37: Ensure a clean ReaderGroup before starting a benchmark
**Problem description**
If the benchmark process is abruptly restarted, it may not be able to close the working `Reader`s and clean up the `ReaderGroup` (e.g., a benchmark runs in a k8 pod and it gets restarted). With the current logic, the benchmark will not be able of restarting the readers as the previous execution did not took care of closing the opened resources (see https://github.com/pravega/pravega-operator/issues/139). 

**Problem location**
`PravegaPerfTest`

Make sure in `PravegaPerfTest` of checking if there are open `Reader`s left from a previous execution, and close them before creating the readers for the next execution.

kmgowda:
Change log description
Currently the reader groups are getting closed once the read compete; This  pull request ensures that reader groups and writers are closed even when the  user terminates the bench-marking.

Purpose of the change
Fixes #37, #41 , #43 

What the code does
To handle ctrl+c ,  the method to call closure of all readers and writer are invoked.
-flush <N>" parameter is added to the pravega benchmark tool.
where N is number of messages to flush per producer.
This option can be used to flush the message by producers while
measuring write performance.

How to verify it
build the tool and issue the writes with -flush <number of events> to verify the issue #41.
run the benchmark tool with consumers and press Ctrl+C while reading ; and re issue the same command to read the data from the beginning again.

Signed-off-by: Keshava Munegowda keshava.munegowda@dell.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ You may obtain a copy of the License at
 The Pravega benchmark tool used for the performance benchmarking of pravega streaming storage cluster.
 This tool performs the throughput and latency analysis for the multi producers/writers and consumers/readers of pravega.
 it also validates the end to end latency. The write and/or read latencies can be stored in a CSV file for later analysis.
-At the end of the performance benchmarking, this tool outputs the 50th, 75th, 95th , 99th and 99.9th latency percentiles.
+At the end of the performance benchmarking, this tool outputs the 50th, 75th, 95th , 99th, 99.9th and 99.99th latency percentiles.
 
 
 ### Prerequisites
@@ -42,7 +42,7 @@ untar the Pravega benchmark tool to local folder
 tar -xvf ./build/distributions/pravega-benchmark.tar -C ./run
 ```
 
-Running Pravega bencmark tool locally:
+Running Pravega benchmark tool locally:
 
 ```
 <dir>/pravega-benchmark$ ./run/pravega-benchmark/bin/pravega-benchmark  -help
@@ -87,8 +87,8 @@ The Pravega benchmark tool can be executed in the following modes:
 ```
 
 ### 1 - Burst Mode
-In this mode, the Pravega benchmark tool pushes/pulls the messages to/from the pravega client as much as possible.
-This mode is used to find the maximum and throughput that can be obtained from the pravega cluster.
+In this mode, the Pravega benchmark tool pushes/pulls the messages to/from the Pravega client as much as possible.
+This mode is used to find the maximum and throughput that can be obtained from the Pravega cluster.
 This mode can be used for both producers and consumers.
 
 ```
@@ -109,8 +109,8 @@ in the case you want to write/read the certain number of events use the -events 
 ```
 
 ### 2 - Throughput Mode
-In this mode, the Pravega benchmark tool pushes the messages to the pravega client with specified approximate maximum throughput in terms of Mega Bytes/second (MB/s).
-This mode is used to find the least latency that can be obtained from the pravega cluster for given throughput.
+In this mode, the Pravega benchmark tool pushes the messages to the Pravega client with specified approximate maximum throughput in terms of Mega Bytes/second (MB/s).
+This mode is used to find the least latency that can be obtained from the Pravega cluster for given throughput.
 This mode is used only for write operation.
 
 ```
@@ -135,8 +135,8 @@ in the case you want to write/read the certain number of events use the -events 
 
 ### 3 - OPS Mode or  Events Rate / Rate Limiter Mode
 This mode is another form of controlling writers throughput by limiting the number of events per second.
-In this mode, the Pravega benchmark tool pushes the messages to the pravega client with specified approximate maximum events per sec.
-This mode is used to find the least latency  that can be obtained from the pravega cluster for events rate.
+In this mode, the Pravega benchmark tool pushes the messages to the Pravega client with specified approximate maximum events per sec.
+This mode is used to find the least latency  that can be obtained from the Pravega cluster for events rate.
 This mode is used only for write operation.
 
 ```
@@ -152,7 +152,7 @@ Note that in this mode, there is 'NO total number of events' to specify hence us
 ```
 
 ### 4 - End to End Latency Mode
-In this mode, the Pravega benchmark tool writes and read the messages to the pravega cluster and records the end to end latency.
+In this mode, the Pravega benchmark tool writes and read the messages to the Pravega cluster and records the end to end latency.
 End to end latency means the time duration between the beginning of the writing event/record to stream and the time after reading the event/record.
 in this mode user must specify both the number of producers and consumers.
 The -throughput option (Throughput mode) or -events (late limiter) can used to limit the writers throughput or events rate.
@@ -166,5 +166,5 @@ The -throughput -1 specifies the writes tries to write the events at the maximum
 ```
 
 ### Recording the latencies to CSV files
-user can use the options "-writecsv  <file name>" to record the latencies of writers and "-readcsv <file name>" for readers.
+User can use the options "-writecsv  <file name>" to record the latencies of writers and "-readcsv <file name>" for readers.
 in case of End to End latency mode, if the user can supply only -readcsv to get the end to end latency in to the csv file.
--- a/README.md
+++ b/README.md
@@ -47,29 +47,31 @@ Running Pravega benchmark tool locally:
 ```
 <dir>/pravega-benchmark$ ./run/pravega-benchmark/bin/pravega-benchmark  -help
 usage: pravega-benchmark
- -consumers <arg>               number of consumers
- -controller <arg>              controller URI
- -events <arg>                  number of events/records if 'time' not
+ -consumers <arg>               Number of consumers
+ -controller <arg>              Controller URI
+ -events <arg>                  Number of events/records if 'time' not
                                 specified;
-                                otherwise, maximum events per second by
-                                producer(s) and/or number of events per
+                                otherwise, Maximum events per second by
+                                producer(s) and/or Number of events per
                                 consumer
+ -flush <arg>                   Number of events/records to flush per
+                                Producer;Not applicable, if both producers
+                                and consumers are specified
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -42,34 +42,36 @@ untar the Pravega benchmark tool to local folder
...
+ -consumers <arg>               Number of consumers
+ -controller <arg>              Controller URI
+ -events <arg>                  Number of events/records if 'time' not
                                 specified;
-                                otherwise, maximum events per second by
-                                producer(s) and/or number of events per
+                                otherwise, Maximum events per second by
+                                producer(s) and/or Number of events per
                                 consumer
+ -flush <arg>                   Number of events/records to flush per
```
tkaitchuck: This explanation is confusing. Should "Flush per producer" maybe be "writer between calls to flush" ?
Also why is it not applicable if there are consumers?

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -172,6 +186,7 @@ public static Test createTest(long startTime, CommandLine commandline, Options o
         final int eventsPerSec;
         final int eventsPerProducer;
         final int eventsPerConsumer;
+        final int flushEventsPerProducer;
```
tkaitchuck: In what sense is this "perProducer"? Isn't it just a configuration parameter for the writer?

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -71,57 +99,105 @@ private String createPayload(int size) {
...
                 recordWrite(payload, stats::recordTime);
-                eCnt.control(i);
             }
             flush();
         }
     }
 
-    private class EventsWriterRW implements Performance {
+
+    private class EventsWriterSleep implements Performance {
```
tkaitchuck: Can these be made static? createBenchmark() is currently creating a loop where this holds onto the outer class which also holds it. This is not just messy but there are going to be a ton of synthetic accessors which I'd like to avoid in a benchmark.

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -71,57 +99,105 @@ private String createPayload(int size) {
...
+                flush();
             }
-            flush();
         }
     }
 
+
     private class EventsWriterTime implements Performance {
 
+        public void benchmark() throws InterruptedException, IOException {
```
tkaitchuck: Here and any other methods that override an interface should have an @Override annotation. 
This avoids the very confusing bug that can occur if the interface changes the method signature, and forgets to update all the impls. If there is a default implementation that will get selected instead of the actual impl and because the code runs without error in can be confusing why something that was supposed to occur didn't.

README.md
kmgowda: I have changed it as :
 -flush <arg>                   Each producer calls flush after writing
                                <arg> number of of events/records; Not
                                applicable, if both producers and
                                consumers are specified

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: ok, I got your point here.
so , I have done here is , instead of creating a class which derives the Performance , i am directly exporting the methods as interface. please have look into these new changes

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: i think its not required now because there are no classes in which benchmark is overridden; instead , i returning returning the function itself as interface

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: yes.. but , since there is variable named  eventsPerProducer , so  i thought of keeping in the same fashion.

tkaitchuck: Right, but that one makes sense. IE if there are 100 events per producer and 10 producers there are 1000 total events.

If there are 10 producers and eventsPerFlush is 10... Then nothing. There is no way to multiply or divide those that gives a number that means anything.

src/main/java/io/pravega/perf/WriterWorker.java
tkaitchuck: EventsWriterTime.benchmark() overrides Performance.benchmark()

kmgowda: now, ts changed as follows :
  private void EventsWriterTime() throws InterruptedException, IOException {
    ....
}

its no more class, its another function assigned as interface Performance.

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: yes your right.. its up to the user/Performance engineer should take care of it. This is introduced because performance team and Raul are having IO issues in PKS; This option was required to observe the performance variations by flushing the events. 
dividing among multiple producers may create more confusions.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -80,6 +80,10 @@ public PerfStats(String action, int reportingInterval, int messageSize, String c
      * Private class for start and end time.
      */
     final private class QueueProcessor implements Callable {
+        final private static int NS_PER_MICRO = 1000;
```
RaulGracia: We don't need many of these constants, please simplify.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -207,6 +222,17 @@ public static Test createTest(long startTime, CommandLine commandline, Options o
                 events = 0;
             }
 
+            if (commandline.hasOption("flush")) {
+                int flushEvents = Integer.parseInt(commandline.getOptionValue("flush"));
+                if (flushEvents > 0) {
+                    flushEventsPerProducer = flushEvents;
+                } else {
+                    flushEventsPerProducer = Integer.MAX_VALUE;
```
RaulGracia: In this case, shouldn't we raise an exception instead of assuming the default value?

src/main/java/io/pravega/perf/PravegaTransactionWriterWorker.java
```diff
@@ -30,7 +31,7 @@
                                    PerfStats stats, String streamName, int eventsPerSec, boolean writeAndRead,
                                    ClientFactory factory, int transactionsPerCommit) {
 
-        super(sensorId, events, secondsToRun, isRandomKey,
+        super(sensorId, events, Integer.MAX_VALUE, secondsToRun, isRandomKey,
```
RaulGracia: Why for transactions flush is set to max?

src/main/java/io/pravega/perf/ReaderWorker.java
```diff
@@ -47,81 +58,78 @@ public Void call() throws InterruptedException, ExecutionException, IOException
...
-                    final long startTime = System.currentTimeMillis();
-                    ret = readData();
-                    if (ret != null) {
-                        stats.recordTime(startTime, System.currentTimeMillis(), ret.length());
-                    }
+
+    public void EventsReader() throws IOException {
+        String ret = null;
+        try {
+            int i = 0;
```
RaulGracia: Using a `for` instead of a `while` you can collapse the variable declaration and increment.

src/main/java/io/pravega/perf/ReaderWorker.java
```diff
@@ -47,81 +58,78 @@ public Void call() throws InterruptedException, ExecutionException, IOException
...
-                        final long endTime = System.currentTimeMillis();
-                        final long startTime = Long.parseLong(ret.substring(0, TIME_HEADER_SIZE));
-                        stats.recordTime(startTime, endTime, ret.length());
-                    }
+
+    public void EventsReaderRW() throws IOException {
+        String ret = null;
+        try {
+            int i = 0;
+            while (i < events) {
```
RaulGracia: Same here about `for` instead of `while`.

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: actually it reduced now  last commit "code cleanup"
ok, lets see the usages :
       final private static int NS_PER_MICRO = 1000;        /* required for NS_PER_MS */
        final private static int MICROS_PER_MS = 1000;     /* required for NS_PER_MS*/
        final private static int NS_PER_MS = NS_PER_MICRO * MICROS_PER_MS;   /* required in further code to calculate the appropriate iterations */
        final private static int PARK_NS = NS_PER_MICRO;   /* I want to keep this as it is for future fine tuning for kafka */

src/main/java/io/pravega/perf/PravegaPerfTest.java
kmgowda: no,  if the user is not supply the parameter -flush , then it normal case is "flush" after all the writes; so , no exception, but we  set to MAX to indicate that flush will be done after writing all the events.

src/main/java/io/pravega/perf/PravegaTransactionWriterWorker.java
kmgowda: for transactions --transactionspercommit is used , so flush is not applicable for transactions.

src/main/java/io/pravega/perf/ReaderWorker.java
kmgowda: increment happens on a condition ; hence i changed it to while

src/main/java/io/pravega/perf/ReaderWorker.java
kmgowda: increment happens on a condition ; hence i changed it to while

kmgowda:
Hi Raul,
   I have tested the same stream for multiple readers and multiple times; it works fine without any issues.

Issue #48, opened 2019-06-26
Use the different Reader group for every invocation


kmgowda:
Currently its observed that if the same readergroup is used ; then if the stream is recreated then reads are failing.
so, its better use the different reader groups for every invocation ; so that even if the stream gets deleted and recreated ; the reader group will always new one for every iteration.

PR #50, opened 2019-07-24
Issues 46 and 48: Add the "-scope" parameter and fix the reader-groups issue

### LINKED ISSUE #46: Scope to be passed in via command line argument
Currently, the scope is hard coded it as 'scope'.

Please parameterize it to allow scope can be any name via command line argument (just as how it handles stream currently).

kmgowda:
**Change log description**  
The "-scope"  parameter is added so that use the pass the name of the scope ; If no scope name is supplied then default scope name "Scope" is used. A different reader group names are used in case of end to end latency. 
The data format is changed from String to Byte Array to avoid the string serialization by Pravega.

**Purpose of the change**
Fixes #46, #48 

**What the code does**
By adding the "-scope" parameter , enabled the user to supply the custom scope names.
Currently reader group delete throws an error in-case of  reader group does not exists; hence different reader groups named used in case of end to end latency.

**How to verify it**
Tested the benchmark tool with and without "-scope" parameter.
Tested the end to end latency repeatedly on bare metal setup and standalone execution.

### DIFF
--- a/src/main/java/io/pravega/perf/PerfStats.java
+++ b/src/main/java/io/pravega/perf/PerfStats.java
@@ -212,19 +212,7 @@ public class PerfStats {
         long totalLatency;
         long maxLatency;
         long totalBytes;
-        ArrayList<LatencyRange> latencyRanges;
-
-        private class LatencyRange {
-            private final int latency;
-            private final int start;
-            private final int end;
-
-            private LatencyRange(int latency, int start, int end) {
-                this.latency = latency;
-                this.start = start;
-                this.end = end;
-            }
-        }
+        ArrayList<int[]> latencyRanges;
 
         LatencyWriter(String action, int messageSize, long startTime) {
             this.action = action;
@@ -242,7 +230,7 @@ public class PerfStats {
             latencyRanges = new ArrayList<>();
             for (int i = 0, cur = 0; i < latencies.length; i++) {
                 if (latencies[i] > 0) {
-                    latencyRanges.add(new LatencyRange(i, cur, cur + latencies[i]));
+                    latencyRanges.add(new int[]{cur, cur + latencies[i], i});
                     cur += latencies[i] + 1;
                     totalLatency += i * latencies[i];
                     count += latencies[i];
@@ -260,10 +248,10 @@ public class PerfStats {
                 percentileIds[i] = (int) (count * percentiles[i]);
             }
 
-            for (LatencyRange lr : latencyRanges) {
+            for (int[] lr : latencyRanges) {
                 while ((index < percentileIds.length) &&
-                        (lr.start <= percentileIds[index]) && (percentileIds[index] <= lr.end)) {
-                    values[index++] = lr.latency;
+                        (lr[0] <= percentileIds[index]) && (percentileIds[index] <= lr[1])) {
+                    values[index++] = lr[2];
                 }
             }
             return values;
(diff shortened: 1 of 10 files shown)


src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -212,19 +212,7 @@ private long windowTimeMS(long time) {
         long totalLatency;
         long maxLatency;
         long totalBytes;
-        ArrayList<LatencyRange> latencyRanges;
```
tkaitchuck: I'm not sure this change is needed.

src/main/java/io/pravega/perf/PravegaStreamHandler.java
```diff
@@ -135,15 +141,25 @@ void recreate() throws InterruptedException, ExecutionException, TimeoutExceptio
...
             rdGroup.resetReaderGroup(rdGrpConfig);
         }
         return rdGroup;
     }
+
+    void deleteReaderGroup() {
+        try {
+            readerGroupManager.deleteReaderGroup(rdGrpName);
+        } catch (InvalidStreamException e) {
+            System.out.println("Reader Group is already deleted");
```
tkaitchuck: In the absence of any context this message would be confusing. Perhaps: "Cannot delete reader group X because it is already deleted" would be better.

src/main/java/io/pravega/perf/ReaderWorker.java
```diff
@@ -54,20 +56,25 @@ private Performance createBenchmark() {
 
     @Override
     public Void call() throws InterruptedException, ExecutionException, IOException {
-        perf.benchmark();
+        try {
+            perf.benchmark();
+        } catch (Exception e) {
+            e.printStackTrace();
```
tkaitchuck: What's the point of printing the stack trace only to rethrow the same one?

src/main/java/io/pravega/perf/WriterWorker.java
```diff
@@ -107,7 +107,12 @@ private Performance createBenchmark() {
 
     @Override
     public Void call() throws InterruptedException, ExecutionException, IOException {
-        perf.benchmark();
+        try {
+            perf.benchmark();
+        } catch (Exception e) {
+            e.printStackTrace();
```
tkaitchuck: Again, why print if it's just going to be rethrown?

src/main/java/io/pravega/perf/PerfStats.java
kmgowda: this is as part of code cleanup.. earlier i had class LatencyRange , now its replaced by Array.

src/main/java/io/pravega/perf/ReaderWorker.java
kmgowda: actually the callables (even though i have them printing in PravegaperTest class) are not printing the stack, so, this is alternative

src/main/java/io/pravega/perf/WriterWorker.java
kmgowda: only for debugging purpose.

src/main/java/io/pravega/perf/PravegaStreamHandler.java
kmgowda: yes.. its changed now..

PR #58, opened 2019-09-06
Issue 57: Add "-fork" option

### LINKED ISSUE #57: Add the option for thread pool executor
**Problem description**
In AWS or cloud deployment where not many CPUs avliable , the thread pool is doing better in terms of burst mode.

**Problem location**
Pravega benchmark tool

add the option "-fork"  , the -fork false uses the conventional threads otherwise it should use fork join.

kmgowda:
**Change log description**  
A new option "-fork" option is added; by default it is set o True; if it set to false, then new fixed thread pool is used.

**Purpose of the change**  
Fixes #57 

**What the code does**  
when user supplies -fork false; conventional thread pool is used; otherwise fork join pool is used.

### DIFF
--- a/src/main/java/io/pravega/perf/PravegaPerfTest.java
+++ b/src/main/java/io/pravega/perf/PravegaPerfTest.java
@@ -284,7 +284,7 @@ public class PravegaPerfTest {
             if (commandline.hasOption("fork")) {
                 fork = Boolean.parseBoolean(commandline.getOptionValue("fork"));
             } else {
-                fork = false;
+                fork = true;
             }
 
             if (commandline.hasOption("throughput")) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -277,6 +281,12 @@ public static Test createTest(long startTime, CommandLine commandline, Options o
                 recreate = producerCount > 0 && consumerCount > 0;
             }
 
+            if (commandline.hasOption("fork")) {
```
claudiofahey: You can simplify this to
`enableConnectionPooling = Boolean.parseBoolean(commandline.getOptionValue("enableConnectionPooling", "true"));
`

kmgowda: yes, its done.. I covered it for all other parameters too.

PR #59, opened 2019-09-06
Issue 56 track unacked; added option to disable connection pooling

claudiofahey:
This PR includes various unrelated updates. Let me know if you need these as separate issues, branches, and PRs.

- Number of unacked events and bytes are now periodically printed
- Added Dockerfile and Kubernetes job
- Removed createScope to allow this to run in Nautilus
- Added option to disable connection pooling for producers
- Statistics are now logged using the slf4j logger so that timestamps are printed

### DIFF
--- a/src/main/java/io/pravega/perf/PravegaWriterWorker.java
+++ b/src/main/java/io/pravega/perf/PravegaWriterWorker.java
@@ -16,25 +16,34 @@ import io.pravega.client.stream.EventStreamWriter;
 import io.pravega.client.ClientFactory;
 import io.pravega.client.stream.impl.ByteArraySerializer;
 import io.pravega.client.stream.EventWriterConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Class for Pravega writer/producer.
  */
 public class PravegaWriterWorker extends WriterWorker {
+    private static Logger log = LoggerFactory.getLogger(PravegaWriterWorker.class);
+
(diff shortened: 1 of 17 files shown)


src/main/java/io/pravega/perf/PravegaWriterWorker.java
```diff
@@ -16,31 +16,41 @@
...
+                EventWriterConfig.builder()
+                        .enableConnectionPooling(enableConnectionPooling)
+                        .build());
     }
 
     @Override
     public long recordWrite(byte[] data, TriConsumer record) {
         CompletableFuture ret;
         final long time = System.currentTimeMillis();
+        record.accept(time, -1, data.length);
```
kmgowda: This will reduce the strength of the tool to pump more data to pravega client; this command method is  used kafka bench-marking too.
implement it as different method something like "recordWriteAck"  and invoke when the user specifies the "-unack true" .

kmgowda: Hi @claudiofahey 
I have raised the pull request : https://github.com/pravega/pravega-benchmark/pull/61 
to extend the tool for kafka bench-marking;  if you are make any changes in the tool ; I suggest you test with kafka too;  This bench-marking tool and existing kafka benchmark tool should yield same results for the single producer.

claudiofahey: I will update the PR once #61 has been merged to master.

claudiofahey:
I will resubmit this PR later.

Issue #60, opened 2019-09-07
Extend the Pravega Benchmark tool for Kafka benchmarking


kmgowda:
**Problem description**
Extend the Pravega benchmark tool for kafka benchmarking too.

**Problem location**
Pravega benchmark tool

**Suggestions for an improvement**
add the option "-kafka"  ,  whenever the user supplies the -kafka true then it should do the kafka bench-marking.

PR #61, opened 2019-09-07
Issue 57 7 & 60: Add "-fork " and "-kafka" options

### LINKED ISSUE #57: Add the option for thread pool executor

kmgowda:
**Change log description**  
A new option "-fork" option is added; by default it is set to True; if it is set to false, then new fixed thread pool is used.
Another new option "-kafka" is added; by default it is set to False; it it is set to true, then kafka bench-marking is conducted.

**Purpose of the change**  
Fixes #57 , #60 

**What the code does**  
when user supplies -fork false; conventional thread pool is used; otherwise fork join pool is used.
when user supplied -kafka true; the kafka bench-marking is conducted.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,7 +31,8 @@ buildscript {
         compile "io.pravega:pravega-client:0.5.0",
                 "io.pravega:pravega-common:0.5.0",
                 "commons-cli:commons-cli:1.3.1",
-                "org.apache.commons:commons-csv:1.5"
+                "org.apache.commons:commons-csv:1.5",
+                "org.apache.kafka:kafka-clients:2.2.0"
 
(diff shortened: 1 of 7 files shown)


maddisondavid:
Are the two options `-fork` and `-kafka` related? On an initial pass of the code it looks like this could be better handled (and digested) as two different PRs

kmgowda:
@maddisondavid 
      yes, it could be handled as two separate PRs..but, since, I was moved out the team; before i left, I just combined all my pending working into single PR.

PR #67, opened 2019-09-17
Issue 66: Added option to disable connection pooling for producers

claudiofahey:


src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -80,6 +80,7 @@ public static void main(String[] args) {
                         "if -1, get the maximum throughput");
         options.addOption("writecsv", true, "CSV file to record write latencies");
         options.addOption("readcsv", true, "CSV file to record read latencies");
+        options.addOption("enableConnectionPooling", true, "enable connection pooling");
```
kmgowda: you need update the readme file also

claudiofahey: README updated. Commit rebased onto latest master.

src/main/java/io/pravega/perf/PravegaWriterWorker.java
```diff
@@ -16,25 +16,32 @@
...
 import io.pravega.client.stream.impl.ByteArraySerializer;
 import io.pravega.client.stream.EventWriterConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Class for Pravega writer/producer.
  */
 public class PravegaWriterWorker extends WriterWorker {
+    private static Logger log = LoggerFactory.getLogger(PravegaWriterWorker.class);
```
RaulGracia: This change may not be needed.

claudiofahey: I believe that you are right that it is not needed now. I needed some extra logging during development so I added it but have since removed the log.info lines. It is useful to have for any future logging needs and doesn't significantly add to runtime performance.

PR #68, opened 2019-09-17
Issue 52: Removed createScope to allow this to run in Nautilus

claudiofahey:
Best-practices for Pravega is to never create a scope. In Nautilus, scopes cannot be created by a normal project user. Outside of Nautilus, scopes can be created using a simple curl command such as
```
curl -X POST -H "Content-Type: application/json" -d '{"scopeName":"examples"}' http://localhost:10080/v1/scopes
```
Signed-off-by: Claudio Fahey <claudio.fahey@dell.com>

### DIFF
--- a/src/main/java/io/pravega/perf/PravegaStreamHandler.java
+++ b/src/main/java/io/pravega/perf/PravegaStreamHandler.java
@@ -66,7 +66,6 @@ public class PravegaStreamHandler {
         this.timeout = timeout;
         this.bgexecutor = bgexecutor;
         streamManager = StreamManager.create(new URI(uri));
-        streamManager.createScope(scope);
         streamconfig = StreamConfiguration.builder().scope(scope).streamName(stream)
                 .scalingPolicy(ScalingPolicy.fixed(segCount))
                 .build();


src/main/java/io/pravega/perf/PravegaStreamHandler.java
```diff
@@ -66,7 +66,6 @@
         this.timeout = timeout;
         this.bgexecutor = bgexecutor;
         streamManager = StreamManager.create(new URI(uri));
-        streamManager.createScope(scope);
```
kmgowda: add an option to create the scope; by default , you can disable it.. this code useful for baremetal deployments... not only limited for nautilus deployment.

maddisondavid: I agree, I would prefer the tool to create the scope if it doesn't exist (as it does currently) BUT provide a way to disable that with an option.

maddisondavid:
This PR has been open for a while and I'd like to get such a change in (i.e. disable scope creation via a flag) ASAP.  I'm happy to do this work if you want to close this PR and I'll put another one up with the correct flag.

I do think it's important that the test is able to create the Scope/Stream if required rather than expecting the user to use an out of band process to create these.  But it is also important that the system allows consumption of an existing stream without trying to create the scope before the test.

claudiofahey:
Closing this PR without merging. It will be replaced by a new PR.

PR #75, opened 2019-09-24
Moved library versions from build.gradle to gradle.properties

claudiofahey:
Signed-off-by: Claudio Fahey <claudio.fahey@dell.com>


RaulGracia:
Already included in #76.

PR #76, opened 2019-09-24
Issue 62: Add Dockerfile and Kubernetes job

### LINKED ISSUE #62: Add Dockerfile and Kubernetes job

claudiofahey:
This includes the commit "Moved library versions from build.gradle to gradle.properties" as it is a prerequisite.

Fixes #62.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,12 +7,16 @@ build
 .classpath
 .DS_Store
 bin/
+ca-certificates/*
+!ca-certificates/README.txt
 classes/
 .settings
 .gradle
 output
 .idea
 target
+lib/*
+!lib/README.txt
 log/
 out/
 .metadata/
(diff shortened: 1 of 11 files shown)


.gitignore
```diff
@@ -7,12 +7,16 @@ build
 .classpath
 .DS_Store
 bin/
+ca-certificates/*
```
RaulGracia: Is this necessary in the current PR?

claudiofahey: In the past, I needed to place the EMC certs in this directory to get anything to build in Docker in our lab. I just tested this today without the EMC certs and it works just fine. I can remove everything related to ca-certificates (in .gitignore and Dockerfile). However, we may want to keep it to make it easy to add CA certificates if they should become necessary in the future.

PR #77, opened 2019-09-24
Issue 64: Statistics are now logged using the slf4j logger so that timestamps are printed

### LINKED ISSUE #64: Use slf4j logger to report statistics
This will cause timestamps to be reported along with the statistics.

claudiofahey:
Statistics are now logged using the slf4j logger so that timestamps are printed.
Added simplelogger.properties.

Fixes #64.

### DIFF
--- a/src/main/resources/simplelogger.properties
+++ b/src/main/resources/simplelogger.properties
@@ -1,0 +1,34 @@
+# SLF4J's SimpleLogger configuration file
+# Simple implementation of Logger that sends all enabled log messages, for all defined loggers, to System.err.
+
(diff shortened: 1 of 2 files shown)


src/main/resources/simplelogger.properties
```diff
@@ -0,0 +1,34 @@
+# SLF4J's SimpleLogger configuration file
```
RaulGracia: Could you add the license header here too, please?

RaulGracia:
@claudiofahey could you please create an issue for this PR? We are enforce to associate PRs to issues.

PR #84, opened 2019-12-22
Issue 91: Fixed formatting of command line examples

maddisondavid:
Fixes the command line examples for each of the different modes so that they are easier to read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ usage: pravega-benchmark
                                      If >= 0 and using transactions,
                                      watermarks will be written on each
                                      commit.
-
+```
 ## Running Performance benchmarking
 
 The Pravega benchmark tool can be executed to
@@ -101,7 +101,7 @@ The Pravega benchmark tool can be executed to
  - write/read the events/records for the specified amount of time
 
 The Pravega benchmark tool can be executed in the following modes:
-```
+
 1. Burst Mode
 2. Throughput Mode
 3. OPS Mode or  Events Rate / Rate limiter Mode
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,8 +114,10 @@ This mode is used to find the maximum and throughput that can be obtained from t
 This mode can be used for both producers and consumers.
 
 ```
```
claudiofahey: The example text only should go within the triple back-ticks. Your change puts the "For example" header in the triple back-ticks. See the rendered file in Github.

maddisondavid: Looking at the rendered file I don't see any `For Example` headers inside the formatted text blocks.  The reason I didn't put the command line example inside a pre-formatted text block is because then it appears as one line in a scrollable box that is difficult to read.

claudiofahey: The formatting is still messed up but I now see that it is not a result of this PR.
For reference, I usually format split long command lines with a backslash separator that can be copied directly to a terminal. For instance:
```
cp /my/long/file1 \
/my/long/file2
```

maddisondavid: Yes, I was going to reformat it like that but didn't want to detract from the examples.

BTW, I don't have write access so I can't merge this.

RaulGracia: @maddisondavid I will merge this PR (in our committer guidelines, it is encouraged that a person different from the author of the PR merges the code).

maddisondavid:
1. Issue 91 opened (my apologies)
2. Examples have now been updated with better formatting and I've also updated all the back ticks as some of the titles were accidentally being "hidden".  The document is now much easier to read.

RaulGracia:
@maddisondavid thanks, looks much better! One minor point: in the `Prerequisites` section of the README, could we delete Gradle? The project adds the wrapper, right? So, I don't think it is a pre-requisite for the user.

PR #88, opened 2020-01-10
Issue 87: Add Throughput Recorder

### LINKED ISSUE #87: Add Throughput Recorder
The tool is currently focused on providing Latency information however throughput is a more important metric for testing of Historical Streams.  The tool should provide throughput information in MB/s as well as percentiles at the end of the test.

**Required Tasks**
- Add ThroughputWriter performance recorder (
    - Counts and records number of bytes recorded in the last second
    - Display percentile information for throughput at end of test
- Add CSVThroughputWriter to record detailed information to CSV file

maddisondavid:
**Change log description**
Adds an extra performance recorder to record Throughput while the test is running

**Purpose of the change**
Fixes #87

**What the code does**
Adds a Throughput writer to the Performance recorder that records the number of bytes every second. This is crucial for performing Historical Reader performance tests. At the end of the test the throughput percentiles are reported in a human readable format.

Also adds two new options for specifying a CSV file of the throughput for each second the test is running:
```
 -readthroughputcsv <arg>            CSV file to record read throughput
 -writethroughputcsv <arg>           CSV file to record write throughput
 -reportingIntervalMillis <arg>      period (in milliseconds) in which
                                     performance will be reported
```
*Recording Throughput*
The CSVThroughput recorder records the Throughput (in MiB/<reporting period>) using the current `TimeWindow`.  If no `reportingIntervalMillis` is specified, the default value of 5 seconds is used.

**How to verify**
If the `readthroughputcsv` or `writethroughputcsv` options are specified then the bytes per reporting period are written to a CSV file for every second the test is running:
```
> pravega-benchmark -controller tcp://localhost:9090 -consumers 1 -scope test -stream test1 -segments 5 -time 10 -readthroughputcsv throughput.csv -reportingIntervalMillis 1000
...
...
> cat throughput.csv
Start,End,Events,Reading Events Throughput,Bytes,Reading MiB Throughput
1578929507066,1578929508069,159641,159163.51,63856400,60.72
1578929508069,1578929509070,159106,158947.05,63642400,60.63
1578929509070,1578929510071,159947,159787.21,63978800,60.95
1578929510071,1578929511072,254130,253876.12,101652000,96.85
1578929511072,1578929512073,185718,185532.47,74287200,70.78
1578929512073,1578929513074,239951,239711.29,95980400,91.44
1578929513074,1578929514075,267084,266817.18,106833600,101.78
1578929514075,1578929515076,263968,263704.3,105587200,100.6
1578929515076,1578929516083,246124,244413.11,98449600,93.24
```


src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -279,7 +418,7 @@ public void printTotal(long endTime) {
             int[] percs = getPercentiles();
 
             log.info(String.format(
-                    "%d records %s, %.3f records/sec, %d bytes record size, %.2f MiB/sec, %.1f ms avg latency, %.1f ms max latency" +
+                    "%,d records %s, %,.3f records/sec, %d bytes record size, %.2f MiB/sec, %.1f ms avg latency, %.1f ms max latency" +
```
maddisondavid: Just to note that I have just remembered that I added commas in several formats to make the numbers easier to read, i.e. `2,684,355 records Writing`.  I'm not sure if anybody is scraping these numbers directly from the output log which would get confused by the commas.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -202,12 +220,133 @@ private long windowTimeMS(long time) {
...
+            this.messageSize = messageSize;
+            this.startTime = startTime;
+            this.throughputs = new ArrayList<>();
+            this.count = 0;
+
+            scheduler = Executors.newSingleThreadScheduledExecutor();
+        }
+
+        public void start() {
+            scheduler.scheduleAtFixedRate(this::recordSecondThroughput, 1, 1, TimeUnit.SECONDS);
```
claudiofahey: You are assuming that this method actually runs exactly every second. I don't think this guarantees execution at a precise enough time. This will result in throughput rates that appear to oscillate when they are actually steady.
Instead, for each period you should calculate the number of ms actually elapsed and use that for the displayed throughput.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -202,12 +220,133 @@ private long windowTimeMS(long time) {
...
+            this.csvFile = csvFile + "-throughput.csv";
+            csvPrinter = new CSVPrinter(Files.newBufferedWriter(Paths.get(csvFile)), CSVFormat.DEFAULT
+                .withHeader("second", action + " Throughput (bytes)"));
+        }
+
+        @Override
+        public void recordThroughput(long seconds, long throughputBytes) {
+            super.recordThroughput(seconds, throughputBytes);
+            try {
+                csvPrinter.printRecord(seconds, throughputBytes);
```
claudiofahey: The most useful data you can have in the CSV file is the unix timestamp `System.currentTimeMillis()` and the total number of bytes written. This allows the user to easily calculate throughput for any period using any type of averaging. It also allows correlating with other metrics.

src/main/java/io/pravega/perf/PerfStats.java
claudiofahey: I like the commas for readability. Check with Tim Butler to see if he scrapes these in his scripts.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -202,12 +220,133 @@ private long windowTimeMS(long time) {
...
+                totalBytesInSecond = 0;
+                seconds++;
+            }
+        }
+
+        public void recordThroughput(long seconds, long throughputBytes) {
+            throughputs.add(totalBytesInSecond);
+        }
+
+        private String[] getPercentiles() {
```
claudiofahey: I don't see the benefit of showing the percentiles of the throughput. There is only 1 data point per second so percentiles are overkill and confusing. I recommend just showing the throughput during the last 1 second every 1 second. If somebody wants to see the throughput during previous intervals, they can just look at previous lines.
To understand the distinction with latency, taking the mean of latencies is dangerous because it loses the most important information (slow operations). So we use percentiles to aggregate the 1000's of latencies per second.
In contract, the throughput that this is calculating is actually the mean throughput for each approximately 1 second interval. Using the mean doesn't lose important information. It is also easy enough to visually inspect the last 10 intervals to see any trends. Using percentiles to aggregate this over all previous time intervals isn't very useful.

maddisondavid: I've now updated the PR by removing the Scheduled Executor and the Throughput Percentiles.

The CSV throughput writer now works by getting a report at the end of every TimeWindow (the interval period is now configurable, previously it was fixed at 5 seconds).  This allows different polling frequencies to be specified for the desired throughput resolution.

maddisondavid: I've also removed the `,` from the Latency output as it turned out some teams were scraping these values from the logs.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -196,17 +203,34 @@ private void record(long bytes, int latency) {
...
         }
 
+        public void lastTime(long time) {
+            this.lastTime = time;
+
+            assert this.lastTime > this.startTime : "Invalid Start and EndTime";
+            this.elapsedSeconds = (this.lastTime - this.startTime) / 1000.0;
+        }
+
+        public double getMBPerSecond() {
```
claudiofahey: Rename this to getMiBPerSecond.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -202,12 +244,40 @@ private long windowTimeMS(long time) {
         }
     }
 
+    /***
```
RaulGracia: Minor: extra `*`.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -202,12 +244,40 @@ private long windowTimeMS(long time) {
...
+     * Perf Recorder that records the number of bytes every second.  Percentiles are reported in human form, i.e. MB/s
```
RaulGracia: Minor: Extra space after dot and missing final dot.

src/main/java/io/pravega/perf/PerfStats.java
```diff
@@ -202,12 +244,40 @@ private long windowTimeMS(long time) {
...
+     */
+    @NotThreadSafe
+    static private class CSVThroughputWriter  {
+        final private CSVPrinter csvPrinter;
+
+        CSVThroughputWriter(String action, String csvFile) throws IOException {
+            csvPrinter = new CSVPrinter(Files.newBufferedWriter(Paths.get(csvFile)), CSVFormat.DEFAULT
+                .withHeader("Start", "End","Events",action+" Events Throughput", "Bytes", action+" MiB Throughput"));
```
RaulGracia: Minor: while we do not added checkstyle in this repo yet, one of the existing rules we apply in other repos is to leave a space after each argument and before/after symbols (like `+`).

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -361,7 +385,7 @@ public static Test createTest(long startTime, CommandLine commandline, Options o
                 } else {
                     action = "Reading";
```
RaulGracia: Maybe these "actions" could be an enum or a constant something for clarity. In any case, this this variable necessary?

RaulGracia:
@maddisondavid I have clones your branch and tried this command:
```
bin\pravega-benchmark -controller tcp://localhost:9090 -consumers 1 -producers 1 -events 100 -size 10 -stream test1 -time 60 -readthroughputcsv read.csv -writethroughputcsv write.csv
```
After the execution of this command, I do see the `read.csv` file but I do not see the `write.csv`. Is this because we only can create CSV for either write or read, or maybe this PR has introduced some problem to the `-writethroughputcsv` option?

The same happens if I use in addition `-readcsv` and `-writecsv`. I only see the "read" latency file being created, and none of the other expected files (read and write throughput, write latency):
```
bin\pravega-benchmark -controller tcp://localhost:9090 -consumers 1 -producers 1 -events 100 -size 10 -stream test1 -time 60 -writethroughputcsv write.csv -readcsv read -writecsv write
``` 
Cannot we have more than one CSV reporter at a time?

maddisondavid:
I wasn't aware of any changes to the existing `readcsv` and `writecsv` options introduced as part of this issue, however I will check the code again and validate.  As you point out, it should be possible to use both at the same time.

maddisondavid:
This seems to have been a change introduced in commit `169fa2bc68d62ccd7484ee053e0a4e864c3a59d3` back in April 2019
https://github.com/pravega/pravega-benchmark/blob/master/src/main/java/io/pravega/perf/PravegaPerfTest.java#L301-L307
```
                writeAndRead = consumerCount > 0;

                if (writeAndRead) {
                    produceStats = null;
                } else {
                    produceStats = new PerfStats("Writing", reportingInterval, messageSize, writeFile, writeThroughputFile);
                }
```
This means that the producerStats will only be created IF we have producers and NO consumers.  It seems to be a deliberate change because it's also repeated when the PerfStats are started, however I don't know why it was changed this way.
https://github.com/pravega/pravega-benchmark/blob/master/src/main/java/io/pravega/perf/PravegaPerfTest.java#L339-L346
```
        public void start(long startTime) throws IOException {
            if (produceStats != null && !writeAndRead) {
                produceStats.start(startTime);
            }
```

src/main/java/io/pravega/perf/PravegaPerfTest.java
maddisondavid: The action is really just used as a label when writing out results.  I believe the variable only exists to make the call to the `PerfStats` constructor easier to read (it wasn't introduced in this PR).  They could be moved to a CONSTANT, but since they're only used once here I'm not sure it would anything.

maddisondavid: These "actions" are just labels used when writing out the results (they weren't added by this PR).  I could change them for constants if you'd prefer, but since they only appear here I'm not sure that really buys anything.

The `action` variable I believe was only introduced in order to make the call to the `PerfStat` constructor cleaner.

maddisondavid:
@RaulGracia I've fixed the conflicts, the change to only write the CSV files for reads or writes was made by a previous commit (not this branch).  If we want to revert it then I think that's best handled by another PR.

In terms of the constants, if you want me to change it I will, but I don't think it would add anything (they're only used as labels).

PR #92, opened 2020-01-12
Issue 89 Add Batch Reader Support

### LINKED ISSUE #89: Add support for reading with BatchClient
Currently the PravegaReader uses the streaming API to read the stream however for Historical read testing we also need to test with the BatchClient. 

**Proposal**
Add an extra `-batchreader` option, that if specified, tells the reader to consume the stream with the BatchClient.  The number of consumers will still be specified and the reader will allocate the segments provided by the BatchClient to the available consumers.

Latency and Throughput metrics would be recorded as they would for any test using the streaming reader.

maddisondavid:
**Change log description**
Adds support for using the BatchReader instead of the StreamingReader within consumers

**Purpose of the change**
Fixes #89

**What the code does**
When reading historic data from a stream data can be read using either the Streaming Reader or the Batch (bounded) Reader.  This adds a new `-batchreaders` option that signifies that the stream should be consumed using the BatchClient.  In this mode all the segments from with a Stream bound are supplied by the Pravega Client and the consumers are free to consume them in any order.

The existing consumer has been renamed `PravegaStreamingReaderWorker` and a new `PravegaBatchReaderWorker` has been introduced.  Each reader is assigned a certain number of the Batch segments with an attempt to give each consumer an equal number of the segments.  If the segment count is lower than the number of specified consumers then the consumer count is reduced so that there is one consumer per Batch segment.

```
 -batchreaders                       signifies the consumers should all be
                                     Batch Readers rather than Streaming
                                     readers
```
**How to verify**
When running with the `-batchreaders` option the test will log the consumer to segment assignments
```
>pravega-benchmark -controller tcp://localhost:9090 -consumers 10 -scope dave -stream test6 -segments 5 -time 20 -batchreaders
...
...
...
2020-01-12 13:16:49:359 +0000 [main] INFO io.pravega.perf.PravegaPerfTest - Limiting To 5 consumers due to small number of segment ranges
2020-01-12 13:16:49:359 +0000 [main] INFO io.pravega.perf.PravegaPerfTest - Segment Assignment 0 -> test6 0(0:2215032)
2020-01-12 13:16:49:359 +0000 [main] INFO io.pravega.perf.PravegaPerfTest - Segment Assignment 1 -> test6 1(0:2175864)
2020-01-12 13:16:49:359 +0000 [main] INFO io.pravega.perf.PravegaPerfTest - Segment Assignment 2 -> test6 2(0:2194224)
2020-01-12 13:16:49:359 +0000 [main] INFO io.pravega.perf.PravegaPerfTest - Segment Assignment 3 -> test6 3(0:2177904)
2020-01-12 13:16:49:359 +0000 [main] INFO io.pravega.perf.PravegaPerfTest - Segment Assignment 4 -> test6 4(0:2189328)
```
As the consumers progress through thier assigned segments ranges they will also log progress:
```
2020-01-12 13:16:49:379 +0000 [ForkJoinPool-2-worker-4] INFO io.pravega.perf.PravegaBatchReaderWorker - id:3 Starting Segment test6, 3(0:2177904)
2020-01-12 13:16:49:379 +0000 [ForkJoinPool-2-worker-3] INFO io.pravega.perf.PravegaBatchReaderWorker - id:2 Starting Segment test6, 2(0:2194224)
2020-01-12 13:16:49:379 +0000 [ForkJoinPool-2-worker-1] INFO io.pravega.perf.PravegaBatchReaderWorker - id:0 Starting Segment test6, 0(0:2215032)
2020-01-12 13:16:49:379 +0000 [ForkJoinPool-2-worker-5] INFO io.pravega.perf.PravegaBatchReaderWorker - id:4 Starting Segment test6, 4(0:2189328)
2020-01-12 13:16:49:379 +0000 [ForkJoinPool-2-worker-2] INFO io.pravega.perf.PravegaBatchReaderWorker - id:1 Starting Segment test6, 1(0:2175864)
...
...
2020-01-12 13:16:49:595 +0000 [ForkJoinPool-2-worker-2] INFO io.pravega.perf.PravegaBatchReaderWorker - id:1 Completed Segment test6, 1(0:2175864)
2020-01-12 13:16:49:595 +0000 [ForkJoinPool-2-worker-2] INFO io.pravega.perf.PravegaBatchReaderWorker - id:1 Completed all assigned assignedSegments
2020-01-12 13:16:49:601 +0000 [ForkJoinPool-2-worker-4] INFO io.pravega.perf.PravegaBatchReaderWorker - id:3 Completed Segment test6, 3(0:2177904)
2020-01-12 13:16:49:601 +0000 [ForkJoinPool-2-worker-3] INFO io.pravega.perf.PravegaBatchReaderWorker - id:2 Completed Segment test6, 2(0:2194224)
2020-01-12 13:16:49:601 +0000 [ForkJoinPool-2-worker-4] INFO io.pravega.perf.PravegaBatchReaderWorker - id:3 Completed all assigned assignedSegments
2020-01-12 13:16:49:601 +0000 [ForkJoinPool-2-worker-3] INFO io.pravega.perf.PravegaBatchReaderWorker - id:2 Completed all assigned assignedSegments
...
...
```

### DIFF
--- a/src/main/java/io/pravega/perf/PravegaBatchReaderWorker.java
+++ b/src/main/java/io/pravega/perf/PravegaBatchReaderWorker.java
@@ -1,0 +1,66 @@
+package io.pravega.perf;
+
+import io.pravega.client.BatchClientFactory;
+import io.pravega.client.batch.SegmentIterator;
+import io.pravega.client.batch.SegmentRange;
+import io.pravega.client.stream.impl.ByteArraySerializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Iterator;
+import java.util.List;
+
+public class PravegaBatchReaderWorker extends ReaderWorker {
+    private static Logger log = LoggerFactory.getLogger(PravegaBatchReaderWorker.class);
+
+    private final BatchClientFactory batchClientFactory;
+    private final Iterator<SegmentRange> assignedSegments;
+
+    private SegmentIterator<byte[]> currentSegmentIterator;
+    private SegmentRange currentRange;
+    private boolean finished;
+
+    PravegaBatchReaderWorker(int readerId, int events, int secondsToRun, long start, PerfStats stats, String readerGrp, int timeout, boolean writeAndRead, BatchClientFactory batchClientFactory, List<SegmentRange> assignedSegments) {
+        super(readerId, events, secondsToRun, start, stats, readerGrp, timeout, writeAndRead);
+        this.batchClientFactory = batchClientFactory;
+
+        this.assignedSegments = assignedSegments.iterator();
+    }
+
+    @Override
+    public byte[] readData() {
+        if (finished) {
+            return null;
+        }
+
+        if (currentSegmentIterator == null || !currentSegmentIterator.hasNext()) {
+            if (currentRange != null) {
+                currentSegmentIterator.close();
+
+                log.info("id:{} Completed Segment {}, {}({}:{})",workerID, currentRange.getStreamName(), currentRange.getSegmentId(), currentRange.getStartOffset(), currentRange.getEndOffset());
+            }
+
+            if (assignedSegments.hasNext()) {
+
+                currentRange = assignedSegments.next();
+                currentSegmentIterator = batchClientFactory.readSegment(currentRange, new ByteArraySerializer());
+
+                log.info("id:{} Starting Segment {}, {}({}:{})",workerID, currentRange.getStreamName(), currentRange.getSegmentId(), currentRange.getStartOffset(), currentRange.getEndOffset());
+            } else {
+                log.info("id:{} Completed all assigned assignedSegments", workerID);
+                currentSegmentIterator = null;
+                finished = true;
+                return null;
+            }
+        }
+
+        return currentSegmentIterator.next();
+    }
+
+    @Override
+    public void close() {
+        if (currentSegmentIterator != null) {
+            currentSegmentIterator.close();
+        }
+    }
+}
--- a/src/main/java/io/pravega/perf/PravegaPerfTest.java
+++ b/src/main/java/io/pravega/perf/PravegaPerfTest.java
@@ -10,8 +10,10 @@
 
 package io.pravega.perf;
 
+import io.pravega.client.BatchClientFactory;
 import io.pravega.client.ClientConfig;
 import io.pravega.client.EventStreamClientFactory;
+import io.pravega.client.batch.SegmentRange;
 import io.pravega.client.stream.ReaderGroup;
 import io.pravega.client.stream.impl.ClientFactoryImpl;
 import io.pravega.client.stream.impl.ControllerImpl;
@@ -23,10 +25,13 @@ import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -43,6 +48,8 @@ import java.util.stream.Stream;
  * Data format is in comma separated format as following: {TimeStamp, Sensor Id, Location, TempValue }.
  */
 public class PravegaPerfTest {
+    private static Logger log = LoggerFactory.getLogger(PravegaPerfTest.class);
+
     final static String BENCHMARKNAME = "pravega-benchmark";
 
     public static void main(String[] args) {
@@ -87,6 +94,8 @@ public class PravegaPerfTest {
         options.addOption("readWatermarkPeriodMillis", true,
                 "If -1 (default), watermarks will not be read.\n" +
                 "If >0, watermarks will be read with a period of this many milliseconds.");
+
+        options.addOption("batchreaders", false, "signifies the consumers should all be Batch Readers rather than Streaming readers");
 
         options.addOption("help", false, "Help message");
 
@@ -186,7 +195,7 @@ public class PravegaPerfTest {
         final boolean recreate;
         final boolean writeAndRead;
         final int producerCount;
-        final int consumerCount;
+        int consumerCount;
         final int segmentCount;
         final int events;
         final int eventsPerSec;
@@ -204,6 +213,7 @@ public class PravegaPerfTest {
         final boolean enableConnectionPooling;
         final long writeWatermarkPeriodMillis;
         final long readWatermarkPeriodMillis;
+        final boolean batchReaders;
 
         Test(long startTime, CommandLine commandline) throws IllegalArgumentException {
             this.startTime = startTime;
@@ -301,6 +311,12 @@ public class PravegaPerfTest {
                 readFile = commandline.getOptionValue("readcsv");
             } else {
                 readFile = null;
+            }
+
+            if (commandline.hasOption("batchreaders")) {
+                batchReaders = true;
+            } else {
+                batchReaders = false;
             }
 
             enableConnectionPooling = Boolean.parseBoolean(commandline.getOptionValue("enableConnectionPooling", "true"));
@@ -403,6 +419,7 @@ public class PravegaPerfTest {
         final PravegaStreamHandler streamHandle;
         final EventStreamClientFactory factory;
         final ReaderGroup readerGroup;
+        final BatchClientFactory batchClientFactory;
 
         PravegaTest(long startTime, CommandLine commandline) throws
                 IllegalArgumentException, URISyntaxException, InterruptedException, Exception {
@@ -426,8 +443,15 @@ public class PravegaPerfTest {
                 }
             }
             if (consumerCount > 0) {
-                readerGroup = streamHandle.createReaderGroup(!writeAndRead);
+                if (batchReaders) {
+                    batchClientFactory = streamHandle.newBatchClientFactory();
+                    readerGroup = null;
+                } else {
+                    batchClientFactory = null;
+                    readerGroup = streamHandle.createReaderGroup(!writeAndRead);
+                }
             } else {
+                batchClientFactory = null;
                 readerGroup = null;
             }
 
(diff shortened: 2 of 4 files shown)


src/main/java/io/pravega/perf/PravegaBatchReaderWorker.java
```diff
@@ -0,0 +1,66 @@
...
+        super(readerId, events, secondsToRun, start, stats, readerGrp, timeout, writeAndRead);
+        this.batchClientFactory = batchClientFactory;
+
+        this.assignedSegments = assignedSegments.iterator();
+    }
+
+    @Override
+    public byte[] readData() {
+        if (finished) {
+            return null;
```
claudiofahey: This will not cause the benchmark process to stop. When all data has been read, the benchmark process will continue to run, reporting 0 events/sec. Is this desired? Maybe you want to throw an exception when the worker completes reading?

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -468,11 +492,43 @@ public void shutdown(long endTime) {
         }
 
         public List<ReaderWorker> getConsumers() throws URISyntaxException {
+            return batchReaders ? getBatchConsumers() : getStreamingConsumers();
+        }
+
+        public List<ReaderWorker> getBatchConsumers() {
```
claudiofahey: A key difference between the batch and streaming readers in this benchmark is that there can be multiple benchmark processes reading together through a reader group. Adding more processes will increase the parallelism and data will not be read multiple times.
The batch reader does not perform this coordination so if one attempts to run multiple reader benchmark processes, each process will read the entire stream.
This behavior is reasonable but it should be documented.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -491,5 +547,33 @@ public void closeReaderGroup() {
...
         }
 
+        /**
+         * Chunks the list of segment ranges between the number consumers.  If the number of segment ranges is
+         * divisible by the number of consumers then each consumer will recieve an equal number of segments, otherwise
+         * some consumers may receive more segments than others.
+         *
+         * @return A list of lists, each list representing the segments assigned to that consumer
+         */
+        private List<List<SegmentRange>> assignSegmentsToConsumers(List<SegmentRange> segmentRanges, int consumers) {
```
claudiofahey: Consider using `com.google.common.collect.Lists.partition` instead of this.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -303,6 +313,12 @@ public static Test createTest(long startTime, CommandLine commandline, Options o
                 readFile = null;
             }
 
+            if (commandline.hasOption("batchreaders")) {
+                batchReaders = true;
+            } else {
+                batchReaders = false;
+            }
+
```
claudiofahey: Simplify to `batchReaders = commandline.hasOption("batchreaders")`.

src/main/java/io/pravega/perf/PravegaPerfTest.java
maddisondavid: That's a good point and it was a subtly in the code that I missed, i.e. that the ReaderGroup name is devised from the steam name itself and therefore there is implicit coordination between multiple distributed processes.

https://github.com/pravega/pravega-benchmark/blob/master/src/main/java/io/pravega/perf/PravegaPerfTest.java#L290-L293
```
            if (recreate) {
                rdGrpName = streamName + startTime;
            } else {
                rdGrpName = streamName + "RdGrp";
            }
```
Obviously, as you point out, this is a bit harder to achieve with the BatchClient as the coordination is basically left up to the processor itself.  I'll document this current restriction and find a way of distributing the processing in another PR.

src/main/java/io/pravega/perf/PravegaPerfTest.java
maddisondavid: Originally I did use Google `Lists.partition` but the problem with that API is that partitions the list based on a specific partition size and NOT (as I'd assumed) by the number of required partitions.  

This lead to getting into calculating the required size of partitions along with handling the cases where the number of SegmentRanges wasn't completely divisible by the number of consumers.  In the end it felt a lot simpler to partition the list as done here.

src/main/java/io/pravega/perf/PravegaBatchReaderWorker.java
maddisondavid: Resolved.  BatchReaders now throw a `WorkerCompletedException` when all assigned events have been completed which stops the test early if there are no more events.

src/main/java/io/pravega/perf/PravegaPerfTest.java
claudiofahey: If you need to distribute the processing (in another PR), consider accepting parameters for the process index and the number of processes. Then each process can sort the available segment ranges and take a deterministic subset.

src/main/java/io/pravega/perf/PravegaBatchReaderWorker.java
```diff
@@ -0,0 +1,66 @@
...
+    public byte[] readData() throws WorkerCompleteException {
+        if (finished) {
+            throw new WorkerCompleteException(workerID);
+        }
+
+        if (currentSegmentIterator == null || !currentSegmentIterator.hasNext()) {
+            if (currentRange != null) {
+                currentSegmentIterator.close();
+
+                log.info("id:{} Completed Segment {}, {}({}:{})",workerID, currentRange.getStreamName(), currentRange.getSegmentId(), currentRange.getStartOffset(), currentRange.getEndOffset());
```
RaulGracia: Minor: missing space after `,`.

src/main/java/io/pravega/perf/PravegaBatchReaderWorker.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            if (assignedSegments.hasNext()) {
+
+                currentRange = assignedSegments.next();
+                currentSegmentIterator = batchClientFactory.readSegment(currentRange, new ByteArraySerializer());
+
+                log.info("id:{} Starting Segment {}, {}({}:{})",workerID, currentRange.getStreamName(), currentRange.getSegmentId(), currentRange.getStartOffset(), currentRange.getEndOffset());
```
RaulGracia: Minor: missing space after `,`.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -88,6 +95,8 @@ public static void main(String[] args) {
                 "If -1 (default), watermarks will not be read.\n" +
                 "If >0, watermarks will be read with a period of this many milliseconds.");
 
+        options.addOption("batchreaders", false, "signifies the consumers should all be Batch Readers rather than Streaming readers");
```
RaulGracia: I see that the rest of command description sentences start by capitals.

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -468,11 +488,43 @@ public void shutdown(long endTime) {
...
+                readers = IntStream.range(0, consumerCount)
+                    .boxed()
+                    .map(i ->
+                        new PravegaBatchReaderWorker(i, eventsPerConsumer,
+                        runtimeSec, startTime, consumeStats,
+                        rdGrpName, TIMEOUT, writeAndRead, batchClientFactory, assignedRanges.get(i))
+                    )
+                    .collect(Collectors.toList());
+            } else {
+                readers = null;
```
RaulGracia: Is this case possible? If the user does not specify `consumers`, we could take as default the number of segments or just `1`. But if the user sets `consumerCount <= 0`, the tool should throw an error even before reaching this point telling the user that the input is incorrect, right?

src/main/java/io/pravega/perf/PravegaPerfTest.java
```diff
@@ -491,5 +543,33 @@ public void closeReaderGroup() {
...
+        /**
+         * Chunks the list of segment ranges between the number consumers.  If the number of segment ranges is
+         * divisible by the number of consumers then each consumer will recieve an equal number of segments, otherwise
+         * some consumers may receive more segments than others.
+         *
+         * @return A list of lists, each list representing the segments assigned to that consumer
+         */
+        private List<List<SegmentRange>> assignSegmentsToConsumers(List<SegmentRange> segmentRanges, int consumers) {
+            List<List<SegmentRange>> results = new ArrayList<>();
+            for (int f=0;f<consumers;f++) {
```
RaulGracia: Minor: Please, leave a space between statements: `(int f=0; f < consumers; f++)`

src/main/java/io/pravega/perf/PravegaBatchReaderWorker.java
```diff
@@ -0,0 +1,66 @@
+package io.pravega.perf;
```
RaulGracia: Missing copyright notice.

src/main/java/io/pravega/perf/WorkerCompleteException.java
```diff
@@ -0,0 +1,21 @@
+package io.pravega.perf;
```
RaulGracia: Missing copyright notice.

src/main/java/io/pravega/perf/ReaderWorker.java
```diff
@@ -100,6 +100,7 @@ public void EventsReaderRW() throws IOException {
                     i++;
                 }
             }
+        } catch(WorkerCompleteException ignore) {
```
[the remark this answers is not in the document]
maddisondavid: No, because the Exception only exists to indicate to the logic that the reader has completed, it's not an actual exception or an error condition.  

The `PravegaBatchReaderWorker` (which is the only worker that throws this exception) will log that it competed everything, so another log would just be redundant.

src/main/java/io/pravega/perf/PravegaPerfTest.java
maddisondavid: This logic follows the same pattern as the streaming consumers.  The problem is that the test doesn't know if consumers are required, it only knows that because `consumers` > 0.  If `consumers` was defaulted to 1 then a test would always have a consumer, which is not desirable.

RaulGracia:
@maddisondavid can you please have a look to the conflicts so we can merge this one?

RaulGracia:
@maddisondavid any updates on this one?

